PR #106, opened 2015-03-02
Show name of the user who created the execution

enykeev:


enykeev:
I'm not particularly fond of how it looks like, but if you think that's ok, feel free to merge.
screen shot 2015-03-02 at 20 45 37

apps/st2-history/template.html
```diff
@@ -113,7 +113,14 @@ <h2 class="st2-flex-table__caption-title">{{ group.period | date:'fullDate':'UTC
                   {{ record.trigger.type }}
                 </span>
               </span>
-              <span ng-switch-default>Manual</span>
+              <span ng-switch-default>
+                <span class="st2-history__column-app-name">
+                  Manual
```
Kami: Btw, why do we say "manual" and not just display the user name? What other possibles values are there besides manual?

enykeev: The execution could either be created by user creating it though API (manually) or by rule (somewhat "automatic"). I'm actually thinking we should also show the application it was called with (direct API call, st2client, web ui, hubot, etc.), but that would require some additional info to be provided (for example, oAuth-like, by issuing an application token).

Kami: I see, thanks for the clarification.

And yeah, I think storing additional context could be useful (e.g. executed via cli / st2web, etc.).

Kami:
LGTM, +1

apps/st2-history/template.html
Kami: To give you an idea - at cloud monitoring, we allowed user to pass special `_who` and `_why` query parameter with each API call - http://docs.rackspace.com/cm/api/v1.0/cm-devguide/content/general-api-info-audits.html

This parameter can then either be populated manually by the user or automatically via the tool (e.g. CLI / st2web) or a combination of those two.

This is very useful for audit and tracking purposes and we should also do something along those lines as well.

enykeev: yeah, cool idea. Could work for us as a first step. Guess the question is how much do we actually care if someone would try to fake this data for whatever reason.

Kami: In general, we shouldn't and don't really care - this information can also be populated by the user (e.g. why have I done this change).

PR #117, opened 2015-03-18
Style refinements

enykeev:


modules/st2-details/style.less
```diff
@@ -272,9 +277,11 @@
 
     display: flex;
 
+    color: #848484;
```
enykeev: Seriously, this gray on orange looks really bad. I don't mind graying out right, inactive part, but for an active one we need to find another way to make a difference for hover... or just don't touch it at all.

enykeev:
Seems like the work with horizontal lines isn't done. You have changed them in some cases but not the other and the margins are still very different across the board.

I do like smaller fonts for descriptions, but you need to make sure fields without descriptions looks alike with the one that have (I'm once again talking about bottom margins).

enykeev:
Another problem that was before but now gets even more apparent is how labels gets truncated.


Bigger labels looks better in list view, but outside of it...

Guess we should move it to the right in Rules.

And yeah, if we are talking about it, we should also fix a difference between left and right paddings in case there is no icon on the label (see &:before's margin).

gmedrano:
>Seriously, this gray on orange looks really bad.

Agreed! Good we clarified it. As we discussed, it will be great if you can separate the font color for each state (active/inactive). Perhaps making overall font "gray" and add color: white to transition. I wasn't able to do it myself.

gmedrano:
>Seems like the work with horizontal lines isn't done. You have changed them in some cases but not the other and the margins are still very different across the board.

This is true. I was trying to highlight separators for each sub-section (Criteria, Trigger, Action Output, etc) by making it brighter than the between-fields separators (cmd, timeout, etc). I'll come up with something less subtle.

gmedrano:
>Another problem that was before but now gets even more apparent is how labels gets truncated.

I think we are using the labeled status only in the Details panel and the small badge (circle) in the list view, no?

gmedrano:
>Bigger labels looks better in list view, but outside of it...

Currently the status indicator is placed under Action Output (History). Should it be moved back to the top "summary" sub-section as in your screenshot?

If so, it can be placed right below the action description line.

enykeev:
Rebased against the latest master. Don't forget to `npm install && bower install`.

enykeev:
Ok, that's not going to work, too much have changed to continue to work in parallel. Merging. @gmedrano create new PR when you have more stuff to share.

PR #121, opened 2015-03-20
Editing Rules

enykeev:
- [x] Editing rules
- [x] Delete the rule
- [x] Create new rule
- [ ] Duplicate existing rule
- [x] Form validation
- [x] Feedback on save
- [ ] Update existing schemes to use validation parameters
- [ ] Support of additional schema properties (minlength, maxlength, format, range, multiples and so on)
- [ ] Review UX
- [ ] Presenting server-side errors (we should preferably avoid them as much as possible)

What else? Feel free to add more ^^


enykeev:
As promised, the list of known issues/planned improvements:
- in edit mode, replace details header (rule name, description and status) with the phrase "Editing the rule" to emphasize the fact edit mode is enabled
- figure a better way to present the result of saving a rule (and reporting an error if it didn't happen). Right now we are using animated label "Saved!" near the save button, but we definitely can do better.
- schema for rule meta (name, description, enabled) is currently hardcoded into UI. Not a big of a problem since we're not going to change it too often, but eventually, we need a way for UI to get the schema from st2common. We would need some kind of `/config` endpoint to get all kinds of different defaults from server side we have no place to put at the moment and schemas might be one of them.
- currently, rule name is validated by `^[\w.-]+$` pattern, but I need a confirmation that it is correct. Also, name is a unique field, so we need to a) add a property to schema definition telling we need an additional check against server side to make sure it's not taken yet and b) start performing this checks (preferably with cheap HEAD requests that returns no body, only status code).
- description field should be textarea instead (since we are expecting it to be rather long), but we don't have suitable form directive at the moment. The other problem is that JSONSchema makes no difference between long and short strings so we would have to pick either one or the other. What makes it even worse is that things like `core.local` cmd param may go all the way from 'short' to 'long' in the matter of one edit and we have no way to replace `input[type-text]` with `textarea` on the fly. Most complete solution would be to create custom form element based on `[contenteditable]` attribute, but that would require a lot of time an effort to make it right.
- combobox directive requires a lot of polish. For one, it needs keyboard support and should limit the number of suggestions it presents at the time. Another thing we need to figure out is how to react when we have a rule that already uses an action we don't currently have in a system.
- pretty much all the actions and rules currently in the system lacks vital validation meta. We need to update them all to make sure we spot as much as possible before we send the rule to server for saving*.
- The response we're currently getting from server isn't helpful enough (and may never be helpful enough) to present to the user what exactly gone wrong. We should use http return codes more extensively (for example, now we are returning 500 when we already have a rule with the same name, but http spec contains code 409 specifically for such case).
- Criteria is another directive that requires a lot of love. It gets thing done at the moment, but it has few weak points related to the nature of the structure we are using to store it (dictionary/object/hash/map, pick the term you like):
  - it is impossible to create a criteria without a name. We are currently cheating by naming it an empty string.
  - it is impossible to define an order for an object properties (as it is impossible to disable automatic sorting). Our 'empty string' criteria always appears on the top of the list, but the button to add a new criteria is on the bottom.
  - it is impossible to rename the criteria in an object. Instead, the new criteria should be created and the old one should be deleted.

  All this problems could be somewhat solved by converting object to an array inside the directive and converting it back to object afterwards using ngModel validation capabilities to make sure we only have unique names, but that requires additional time and effort. What bothers me is that we still have no idea how exactly are we going to implement 'OR' in rules. We may end up switching to the entirely new format for criteria potentially solving the problem without additional effort on our side.
- switching to another rule/app during rule editing should result in confirmation dialog asking you whether you really want to cancel.
- if rule is deleted from the system, it should be also deleted from `/opt/stackstorm` folder. Otherwise, the rule would be restored after next system restart. It was kind of handy during development, but it should stop now.

\* In general, I'd say we should do everything we could so UI could catch the problem before sending. It includes complete schema-based validation and cheap HEAD requests to make sure unique parameters are not yet taken (think of rule name as an example). For cases where we for some reason failed to predict server would reject the request, report should be clean and machine-readable (first of all, it should include proper return code). We neither have place to put complete error message nor the way to parse something like an output returned by python jsonschema validator.

PR #126, opened 2015-04-14
Reduce the size of production assets

enykeev:
By removing some unnecessary packages and replacing poorly packed hljs with Lea Verou's Prism, we've managed to reduce the size of `components.js` in half.

There is still an opportunity to improve on that result by using pre-uglified versions of packages (mainly, angular and lodash), but that's a minor improvement (estimated around 5-7%).


Kami:
LGTM

PR #136, opened 2015-04-20
Criteria suggestions

enykeev:


modules/st2-criteria/directive.js
```diff
@@ -8,30 +8,93 @@ angular.module('main')
...
       templateUrl: 'modules/st2-criteria/template.html',
-      link: function postLink(scope) {
+      link: function postLink(scope, element, attrs, ctrl) {
         scope.ngModel = scope.ngModel || {};
 
-        scope.remove = function (key) {
-          delete scope.ngModel[key];
+        scope.typeSpec = {
+          enum: {
+            'matchregex': 'Matches Regex',
```
Kami: Nice. Eventually we could also add a "meta" API for retrieving available criteria comparision operators.

enykeev: yep, that's the plan. Eventually, we can start retrieving a lot of config-like stuff from the API: model schemes to build panels like Action or Rule Details, actionreporters for new runner types, langpacks, custom jsonschema extensions. This would really help us if/when we go *aaS route or at least start providing single UI deployment for people to use with their local st2 instances.

PR #137, opened 2015-04-23
Feedback for v0.9

enykeev:


enykeev:
Ok, so lets recap, big pieces that still bother me:

- Bad error reporting (in general, but especially on rule manipulations). We can use Details header for that purpose (temporarily replace name and description with error message) or we can add global notification system (basically a popup in top right corner). The latter is more universal, but also a bigger one of two. A viable option is to put a problem on the backlog for 0.9.

- Criteria key uniqueness. Right now you can create as much criteria with the same key as you want, but only one would end up in a rule we're going to send to the server. And the more manipulation you are doing with criteria in one go, the harder it is to predict which one it would be. To solve that problem properly, we need to review several form components at once to to pass the necessary scope variables in, make sure we get right validation data out and manually (since we still have no tests) check that this changes doesn't affect the behavior of the components anywhere else in the project.

There is also at least one small piece related to previous feedback:

- Need to position buttons on top in rule creation dialog. I don't see a good way to do that at the moment.

Am I missing something?

enykeev:
STORM-1259

enykeev:
Another thing we forgot to restyle

PR #138, opened 2015-04-24
RFR: Use ref to handle rules as opposed to id

lakshmi-kannan:
Dependent on https://github.com/StackStorm/st2/pull/1367

I hope I caught all the `id` to ref changes for the `rules`.

### DIFF
--- a/apps/st2-rules/controller.js
+++ b/apps/st2-rules/controller.js
@@ -20,10 +20,10 @@ angular.module('main')
         url: '/new'
       })
       .state('rules.general', {
-        url: '/{id:\\w+}/general?edit'
+        url: '/{ref:\\w+}/general?edit'
       })
       .state('rules.code', {
-        url: '/{id:\\w+}/code?edit'
+        url: '/{ref:\\w+}/code?edit'
       })
 
       ;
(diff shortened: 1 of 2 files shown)


lakshmi-kannan:
I have no idea how we the UI works with rules without this PR :). 
cc: @enykeev

Kami:
@lakshmi-kannan IIRC, You can also use ID (or ref) when working with rules endpoint so it should probably also work with the ID, right?

lakshmi-kannan:
@Kami You are right but @enykeev suggested I switch to ref. I don't remember why. I am just surprised this hasn't been merged :P.

Kami:
It's fine, I think ref is better as well :)

manasdk:
Why is `ref` better than `id`? AFAICT there is an index on `id` and none for `ref` so it should be better to use `id` if available.

Kami:
@manasdk It's more user-friendly and consistent with other views (e.g. actions).

And with user-friendly I mean when it's used inside the URL when retrieving details for a single rule. Besides the URL, I don't care and I'm also fine with using ID internally in other places.

Kami:
Also, re index - there will be index once my rbac branch is done and we merged it (I've updated all the resource models to explicitly store "ref" in the DB and added index on it).

enykeev:
I've rebased it locally and it seems like the code is still perfectly valid. Merging.

PR #144, opened 2015-04-28
Smart text field

### LINKED ISSUE #129: Smart text field
The `st2-form-input` directive we're using right now for `type: 'string'` fields isn't good enough when the text inside it is bigger than the field itself (which is undesirable yet not unlikely situation when we're dealing with user's existing scripts).

The main problem is that it is hard to scroll through the input element, especially when it is disabled (as in case of history records where user may not even realize he still can scroll this field)

New directive should automatically resize vertically depending on the length of the string. As soon as the line no longer fits in one line, it should be wrapped to the next line. Whether it should break the line by words or by characters is up for discussion.

It should inherit from ngModel to support validation and ngModelOptions out the box.

New directive is for `type: 'string'` fields only so no type conversion is necessary. On the other hand, since it is now a highly specialized field, it should support all the JSON schema properties of the [string type](http://spacetelescope.github.io/understanding-json-schema/reference/string.html) (min\maxLength, pattern and filters in the strict compliance with the RFCs provided) along with generic metadata (name, description, default, required, immutable).

There is also should be a possibility to define minimum and maximum number of rows allowed for the particular instance of the directive. This is mostly required for cases like `core.sendmail` body field when we pretty certain it would likely take more than four-five lines of text and we want to limit it to keep the form together.

From implementation standpoint, I believe it should be done with [contenteditable] element. I expect it would require a lot more effort to make UX work (keyboard support and all that stuff) than to just pick textarea, although I have a feeling textarea won't be enough for everything we want to achieve and we may spend even more time trying to hack around its limitations. Although, I might be wrong so it is up to the author to pick the right way.

alex-shnayder:
Resolves #129

### DIFF
--- a/modules/st2-auto-form/modules/st2-form-text-field/directive.js
+++ b/modules/st2-auto-form/modules/st2-form-text-field/directive.js
@@ -1,0 +1,196 @@
+;(function(angular, undefined) {
+  'use strict';
+
+  angular.module('main')
+    .directive('st2FormTextField', [
+      '$window',
+      function (window) {
+        return {
+          restrict: 'C',
+          require: 'ngModel',
+          scope: {
+            'spec': '=',
+            'options': '=',
+            'ngModel': '=',
+            'disabled': '='
+          },
+          templateUrl: 'modules/st2-auto-form/modules/st2-form-text-field/template.html',
+          link: function (scope, $element, attrs, ctrl) {
+            var textarea = $element[0].querySelector('.st2-auto-form__field'),
+                $textarea = angular.element(textarea);
+
+            scope.name = ctrl.$name;
+
+            if (!textarea || textarea.nodeName !== 'TEXTAREA' ||
+                !window.getComputedStyle || $textarea.data('st2FormTextField')) {
+              return;
+            }
+
+
+            var $container = $textarea.parent(),
+                $clone = angular.element('<textarea class="st2-auto-form__field-clone" tabindex="-1">'),
+                clone = $clone[0],
+                initialStyle = window.getComputedStyle(textarea),
+                initialResize = initialStyle.getPropertyValue('resize'),
+                boundAdjust = adjust.bind(window, window, $textarea, $clone),
+                adjustAndValidate = function() {
+                  boundAdjust();
+
+                  setTimeout(function() {
+                    $container.toggleClass('st2-auto-form__field-container--invalid', $textarea.hasClass('ng-invalid'));
+                  }, 0);
+                },
+                isTouched = false;
+
+            $clone.data('st2FormTextField', true);
+            $element.append(clone);
+            $textarea
+              .css({
+                'resize': (initialResize === 'none' || initialResize === 'vertical') ? 'none' : 'horizontal'
+              })
+              .data('st2FormTextField', true)
+              .on('focus blur', function() {
+                $container.toggleClass('st2-auto-form__field-container--focused');
+
+                if (isTouched) {
+                  $container.addClass('st2-auto-form__field-container--touched');
+                }
+
+                isTouched = true;
+              })
+
+              // Prevent textarea scroll bubbling
+              .on('mousewheel DOMMouseScroll', function(e) {
+                var delta = -e.wheelDeltaY || e.detail;
+
+                if (textarea.scrollHeight > textarea.clientHeight &&
+                     (delta < 0 && !textarea.scrollTop ||
+                      delta > 0 && textarea.scrollTop + textarea.clientHeight === textarea.scrollHeight)
+                   ) {
+                  e.preventDefault();
+                }
+              });
+
+            scope.$watch(function() {
+              return ctrl.$modelValue;
+            }, adjustAndValidate);
+
+            textarea.addEventListener('input', adjustAndValidate, false);
+
+            // Oninput in IE9 doesn't capture character deletion,
+            // so we use onkeyup to overcome this
+            if ('onpropertychange' in textarea && 'oninput' in textarea) {
+              textarea.addEventListener('keyup', boundAdjust, false);
+            }
+
+            scope.$on('$destroy', function() {
+              $clone.remove();
+            });
+
+            scope.$watch('minRows', function(minRows) {
+              $textarea.minRows = window.parseInt(minRows, 10);
+              boundAdjust();
+            });
+
+            scope.$watch('maxRows', function(maxRows) {
+              $textarea.maxRows = window.parseInt(maxRows, 10);
+              boundAdjust();
+            });
+
+            setTimeout(boundAdjust, 0);
+          }
+        };
+      }
+    ]);
+
+
+  function getDimensions(window, $textarea, computedStyle) {
+    var dimensions = {
+          width: window.parseInt(computedStyle.getPropertyValue('width'), 10),
+          height: window.parseInt(computedStyle.getPropertyValue('height'), 10),
+          minHeight: window.parseInt(computedStyle.getPropertyValue('min-height'), 10),
+          maxHeight: window.parseInt(computedStyle.getPropertyValue('max-height'), 10),
+          outerHeight: 0
+        },
+        lineHeight = window.parseInt(computedStyle.getPropertyValue('line-height'), 10);
+
+    if (computedStyle.getPropertyValue('box-sizing') === 'border-box') {
+      dimensions.outerHeight =
+        window.parseInt(computedStyle.getPropertyValue('padding-top'), 10) +
+        window.parseInt(computedStyle.getPropertyValue('padding-bottom'), 10) +
+        window.parseInt(computedStyle.getPropertyValue('border-top-width'), 10) +
+        window.parseInt(computedStyle.getPropertyValue('border-bottom-width'), 10);
+    }
+
+    dimensions.minHeight = dimensions.minHeight > 0 ? dimensions.minHeight : (lineHeight + dimensions.outerHeight);
+    dimensions.maxHeight = dimensions.maxHeight > 0 ? dimensions.maxHeight : Infinity;
+
+    if ($textarea.minRows) {
+      dimensions.minHeight = window.Math.max(dimensions.minHeight,
+                                             $textarea.minRows * lineHeight + dimensions.outerHeight);
+    }
+
+    if ($textarea.maxRows) {
+      dimensions.maxHeight = window.Math.min(dimensions.maxHeight,
+                                             $textarea.maxRows * lineHeight + dimensions.outerHeight);
+    }
+
+    return dimensions;
+  }
+
+  function imitate(window, source, target, computedStyle) {
+    var propertiesToCopy = [
+          'box-sizing',
+          'width',
+          'padding-top',
+          'padding-right',
+          'padding-bottom',
+          'padding-left',
+          'border-top-width',
+          'border-right-width',
+          'border-bottom-width',
+          'border-left-width',
+          'font-family',
+          'font-size',
+          'font-weight',
+          'font-style',
+          'letter-spacing',
+          'line-height',
+          'text-transform',
+          'word-spacing',
+          'text-indent'
+        ],
+        styleString = '';
+
+    angular.forEach(propertiesToCopy, function(property) {
+      styleString += property + ':' + computedStyle.getPropertyValue(property) + ';';
+    });
+
+    target.setAttribute('style', styleString);
+    target.value = source.value;
+  }
+
+  function adjust(window, $textarea, $clone) {
+    var computedStyle = window.getComputedStyle($textarea[0]),
+        dimensions = getDimensions(window, $textarea, computedStyle),
+        currentOverflow = computedStyle.getPropertyValue('overflow-y'),
+        newOverflow = 'hidden',
+        newHeight;
+
+    imitate(window, $textarea[0], $clone[0], computedStyle);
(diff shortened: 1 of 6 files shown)


bower.json
```diff
@@ -20,7 +20,8 @@
     "angular-sanitize": "~1.3.4",
     "uri.js": "~1.14.2",
     "angular-busy": "~4.1.2",
-    "prism": "f658c05"
+    "prism": "f658c05",
+    "angular-value-format": "~0.1.0"
```
enykeev: I see a clear benefit of separating such functionality. Re-usability and all. What bothers me is the number of my PRs hanging for half a year in other projects we depend on. This module is simple enough so it would not require a lot of changes, though, I have to ask, can you guarantee you are going to support it for at least a couple of years? Because if not, I would rather prefer we would make this a part of st2web from day one.

modules/st2-auto-form/modules/st2-form-text-field/directive.js
```diff
@@ -0,0 +1,29 @@
...
+      link: function (scope, element, attrs, ctrl) {
+        scope.name = ctrl.$name;
+        
+        ctrl.$render = function () {
+          scope.rawResult = ctrl.$viewValue;
+        };
+        
+        scope.$watch('rawResult', function (rawResult) {
+          ctrl.$setViewValue(rawResult);
+        });
```
enykeev: Since we're not doing any kind of transformation here, we can just map scope.ngModel directly to textarea ng-model attribute. Though, you may hold on for a second with that one since I think we may want to make this directive a bit more complex.

modules/st2-auto-form/modules/st2-form-text-field/template.html
```diff
@@ -0,0 +1,29 @@
...
+    ng-required="spec.required"
+    ng-model="rawResult"
+    ng-model-options="options || {}"
+    ng-pattern="spec.pattern"
+    ng-disabled="disabled"
+    minlength="{{spec.minLength}}"
+    maxlength="{{spec.maxLength}}"
+    btsx-value-format="{{spec.format}}"
+    st2-elastic-field-min-rows="{{spec.minRows}}"
+    st2-elastic-field-max-rows="{{spec.maxRows}}">
```
enykeev: For one, I don't think that min/maxRows should be a part of the spec. Spec mostly defines the data and its constraints, not representation. Instead, we should probably make this parameters scope attributes of st2FormTextField and hardcode them for now at this level.

modules/st2-elastic-field/directive.js
```diff
@@ -0,0 +1,177 @@
+'use strict';
+angular.module('main')
+  .directive('st2ElasticField', [
```
enykeev: Another thing with `st2ElasticField` is that I'm starting to wonder, if we're making it a separate directive, where else can we use it? From line 115, it seems like we're only limited by another textarea element and that's it.

At the moment, I don't see another use for this directive and if you also don't have anything in mind, I suggest we move the whole logic into `st2FormTextField` directive.

bower.json
alex-shnayder: Well, to be honest, I can't. I'll remake it into an st2 directive then.

modules/st2-elastic-field/directive.js
```diff
@@ -0,0 +1,177 @@
...
+          dimensions.outer = {
+            width:  window.parseInt(computedStyle.getPropertyValue('padding-left'), 10) +
+                    window.parseInt(computedStyle.getPropertyValue('padding-right'), 10) +
+                    window.parseInt(computedStyle.getPropertyValue('border-left-width'), 10) +
+                    window.parseInt(computedStyle.getPropertyValue('border-right-width'), 10),
+            height: window.parseInt(computedStyle.getPropertyValue('padding-top'), 10) +
+                    window.parseInt(computedStyle.getPropertyValue('padding-bottom'), 10) +
+                    window.parseInt(computedStyle.getPropertyValue('border-top-width'), 10) +
+                    window.parseInt(computedStyle.getPropertyValue('border-bottom-width'), 10)
+          };
```
enykeev: I'm using native js rarely so I might be missing something important here, tell me if I do. This part seems very complicated to me... can we replace this dimension calculations with something a little bit simplier, like offsetWidth\Height, instead of relying on computed styles? Is there are some pitfalls I don't know about?

modules/st2-elastic-field/directive.js
```diff
@@ -0,0 +1,177 @@
...
+        require: 'ngModel',
+        link: function (scope, $textarea, attrs, ngModel) {
+          if ($textarea[0].nodeName !== 'TEXTAREA' || !window.getComputedStyle || $textarea.data('elasticField')) {
+            return;
+          }
+          
+          
+          var $window = angular.element(window),
+              textarea = $textarea[0],
+              $clone = angular.element('<textarea class="st2-elastic-field-clone" tabindex="-1">'),
```
enykeev: Also wonder, can we do it without introducing additional element, dunno, for example by making parent container `overflow: hidden` and then comparing the height of this container to the height of the textarea inside it. It is OK if you tell me we can't do that, or that the solution would become too complex, or that there is another valid reason to do it this way. I just interested to know what other options do we have.

In any way, if we decide we need additional element, lets at least create it inside `st2FormTextField` and not pollute body with it.

modules/st2-elastic-field/directive.js
alex-shnayder: What are the pros of such a move? I see at least one benefit of the separation: it's future-proof. But I don't see any cons.

modules/st2-elastic-field/directive.js
```diff
@@ -0,0 +1,177 @@
...
+      function adjust($textarea, $clone, force) {
+        if ($textarea.isBeingAdjusted && !force) {
+          return;
+        }
+        
+        var dimensions = getDimensions($textarea),
+            overflow = 'hidden',
+            cloneHeight;
+        
+        $textarea.isBeingAdjusted = true;
```
enykeev: Why do we need this flag here? Once again, I might be missing something, but JS is single threaded and we can't possibly expect another function to access this element between this line and 108 where we switching it back to false.

modules/st2-elastic-field/directive.js
```diff
@@ -0,0 +1,177 @@
...
+          angular.element(document.body).append(clone);
+
+
+          textarea.addEventListener('input', boundAdjust, false);
+          
+          // Oninput in IE9 doesn't capture character deletion,
+          // so we use onkeyup to overcome this
+          if ('onpropertychange' in textarea && 'oninput' in textarea) {
+            textarea.addEventListener('keyup', boundAdjust, false);
+          }
```
enykeev: Just as the heads up, at that point we don't really care about IE9... or 10. It is nice to have indeed, but if in the future it might take you too much time, just skip it.

modules/st2-elastic-field/directive.js
```diff
@@ -0,0 +1,177 @@
...
+
+          $window.bind('resize', boundForcedAdjust);
```
enykeev: The way this directive works, this function would be useful only in case elements width would change because of window resize. In our case, form elements always reside within fixed width container by design, so it is not required, for now.

If you decide to keep it, that's fine, but then I wonder whether it might introduce a flood of recalcs and repaints? Should not we use requestAnimationFrame or something like that here? I'm usually avoiding direct DOM access so I don't have to think about that, letting angular handle that kind of problems for me.

enykeev:
There is also a small issue with scroll, it sticks to the right border of the block.

It is basically how webkit scrolls work, you can't see a margin on a scroll itself so you have to add padding to its outer element.

modules/st2-elastic-field/directive.js
alex-shnayder: I agree, this part looks pretty monstrous. I think it might be possible to improve it using offset/clientHeight and the already calculated height, but I need to investigate a little into browsers' behavior.

enykeev:
Additionally, there is weird UX behavior: when you scroll inside the block and you reach the end, the parent block starts scrolling. That's a huge piece and it might deserve its own ticket, but I believe we need to solve that eventually.

modules/st2-elastic-field/directive.js
enykeev: There is probably none, it is just my fear that with the number of modules we have we may want to start writing our very own package manager by that point =) Just trying to avoid creating new modules, at least until we get rid of ones we don't need anymore.

enykeev:
Yeah, forgot to mention the most important thing.

I like it! Nicely done, @alex-shnayder. Besides the small visual moments, it looks and works exactly as it should, no questions here. I'm ready to merge it any moment (well, we're in the release cycle right now, so "any moment" = "starting next week" =).

What I want now is to show you what kind of questions I'm asking myself in such situations and also figure out where can we go from there. I still want you to fix some of this things, one we can agree on.

modules/st2-elastic-field/directive.js
alex-shnayder: We use scrollHeight to determine the height of the text in a textarea and then we set the height of the textarea itself equal to it. It would be a perfect property for the case if not one flaw: it is never less than textarea's height. That's why we use a hidden clone with zero height.
I don't know of any other approaches to the problem (except contenteditable, which is a whole another story). And I don't see how container's `overflow: hidden` can help here.

enykeev: Yeah, now I see. There is probably a way to set height to 0 on an active element, get its dimensions, set it back to the one it should be and hope it won't trigger a repaint, but that's too much uncertainty to care. Lets just move hidden element inside `st2-form-text-field`.

modules/st2-elastic-field/directive.js
alex-shnayder: Well, now I see the benefit. By moving this logic to `st2FormTextField` we can make it more case-specific (like stop polluting the body).

modules/st2-elastic-field/directive.js
alex-shnayder: It is indeed a very simple approach that can lead to small perfomance issues on window resize. Since we are moving this logic to `st2FormTextField`, I guess we may remove it.

modules/st2-elastic-field/directive.js
alex-shnayder: I guess you are right. The flag is there from the library the code is based on. I wanted to review the flag after I rewrite the code, but forgot about it.

enykeev:
Please, stage index.html too. It is auto-generated, but I prefer to have the latest version always available in repo to see what files it's going to load.

alex-shnayder:
Thank you :)
I'll wrap the textarea in a similar-looking div with padding to unstick the scrollbar. The weird behavior is native, but I see why you'd want to get rid of it. I recall dealing with it long time ago and I think it was pretty easy (stopping propagation, as you said, of a textarea scroll event, probably).
I have another task that will take a couple of days. After that, I'll fix all the issues we agreed on and update the pull request before the next week begins. Is it okay?

enykeev:
Yeah, totally fine by me. As I said, we ourselves are a little bit busy this week.

alex-shnayder:
I've updated the pull request, resolving all the issues we talked about. A couple of them required more effort than I anticipated, though. 
Regarding the trailing whitespace, are you sure you want me to remove the empty lines at the EOF? I tried not to add any, but some files already had them.

modules/st2-auto-form/modules/st2-form-text-field/directive.js
```diff
@@ -0,0 +1,196 @@
...
+        },
+        lineHeight = window.parseInt(computedStyle.getPropertyValue('line-height'), 10);
+
+    if (computedStyle.getPropertyValue('box-sizing') === 'border-box') {
+      dimensions.outerHeight =
+        window.parseInt(computedStyle.getPropertyValue('padding-top'), 10) +
+        window.parseInt(computedStyle.getPropertyValue('padding-bottom'), 10) +
+        window.parseInt(computedStyle.getPropertyValue('border-top-width'), 10) +
+        window.parseInt(computedStyle.getPropertyValue('border-bottom-width'), 10);
+    }
```
alex-shnayder: Didn't come up with a viable solution using client/offsetHeight, but at least removed the unnecessary part with outer width.

enykeev:
Thanks, Alex. I'll take a look as soon as I'm done with my current assignment.

> Regarding the trailing whitespace, are you sure you want me to remove the empty lines at the EOF? I tried not to add any, but some files already had them.

I want every file to end with exactly one `\n` so github wouldn't show red icon. What editor are you using? Github Atom does that with `whitespace` core package OTB.

alex-shnayder:
Oh, that I can do. Updated the commit once again with the trailing `\n` added.

modules/st2-auto-form/modules/st2-form-text-field/directive.js
enykeev: So what went wrong with client/offsetHeight? I have no clear understanding on the topic so may be you can explain to me why we can't use them in that particular case, say, for an educational purposes.

alex-shnayder: The goal here is to calculate the vertical padding + border. We could subtract the inner height from the full height, but there is no one-step way to get the former since both offsetHeight and clientHeight include padding in their return value. So, we have to use the computed style for padding. `border-top + border-bottom` can be replaced by `offsetHeight - clientHeight`, but it doesn't make a big difference (shorter, but also less straightforward).

enykeev:
Ok, merging. We'll see what we've missed as soon as bug reports starts piling up. If they ever.

PR #150, opened 2015-05-06
Remember the groups collapsed

enykeev:
Reported by @manasdk.

Now any group collapsed by clicking the header stay collapsed during the whole session. The setting doesn't persist through tab reload.


enykeev:
@emedvedev please, rebase\reimplement

PR #156, opened 2015-06-08
Execution rerun

### LINKED ISSUE #141: I'd like to re-run an action via UI just like CLI

enykeev:
Closes #141
Depends on StackStorm/st2client.js#26

### DIFF
--- a/apps/st2-history/controller.js
+++ b/apps/st2-history/controller.js
@@ -22,6 +22,9 @@ angular.module('main')
       .state('history.code', {
         url: '/{id:\\w+}/code?page&status&action&rule&trigger&trigger_type'
       })
+      .state('history.rerun', {
+        url: '/{id:\\w+}/rerun?page&status&action&rule&trigger&trigger_type'
+      })
 
       ;
 
@@ -217,7 +220,7 @@ angular.module('main')
 
         _.assign(node, record);
 
-        if ($scope.record.id === record.id) {
+        if ($scope.record && $scope.record.id === record.id) {
           _.assign($scope.record, record);
         }
 
@@ -252,6 +255,46 @@ angular.module('main')
         });
       }
     };
+
+    $scope.rerun = (function () {
+      var rerun = $scope.$new();
+
+      rerun.metaSpec = {
+        type: 'object',
+        properties: {
+          ref: {
+            type: 'string',
+            name: 'Action'
+          }
+        }
+      };
+
+
+      rerun.open = function () {
+        $scope.$root.state.go('^.rerun', {id: $scope.record.id});
+        rerun.payload = _.clone($scope.payload);
+        rerun.actionSpec = $scope.actionSpec;
+      };
+
+      rerun.cancel = function () {
+        $scope.$root.state.go('^.general', {id: $scope.record.id});
+      };
+
+      rerun.submit = function () {
+        st2api.client.executions.repeat($scope.record.id, {
+          parameters: rerun.payload
+        }).then(function (record) {
+          $scope.$root.state.go('^.general', {id: record.id});
+        }).catch(function (error) {
+          $scope.rerunform.err = true;
+          $scope.$apply();
+          $scope.rerunform.err = false;
+          console.error(error);
+        });
+      };
+
+      return rerun;
+    })();
 
   })
 
--- a/apps/st2-history/template.html
+++ b/apps/st2-history/template.html
@@ -204,6 +204,11 @@
 
     <div class="st2-details__toolbar">
 
+      <input type="button"
+          class="st2-forms__button st2-forms__button--small st2-details__toolbar-button"
+          ng-click="rerun.open()"
+          value="Rerun"/>
+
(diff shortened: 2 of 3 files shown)


apps/st2-history/controller.js
```diff
@@ -253,6 +256,46 @@ angular.module('main')
...
+            type: 'string',
+            name: 'Action'
+          }
+        }
+      };
+
+
+      rerun.open = function () {
+        $scope.$root.state.go('^.rerun', {id: $scope.record.id});
+        rerun.payload = _.clone($scope.payload);
```
Kami: So we clone all the parameters on the client side and send all of them to the server (or we just send the overrides)?

If we do the former I'm fine fine that since it probably keeps the client side code simpler (we don't need to figure what has changed and the end result should be the same anyway).

enykeev: We're cloning the parameters execution has been ran with. We're doing it so user would understand what to expect. If we don't do that, UI would show either an empty field (and will left user wondering what exactly do he want to change here) or `default` value from action schema (which would be misleading).

So, say we have an execution of `core.local` with parameters `{cmd: 'ls -lah', cwd: '/'}`. The `rerun.payload` would be the same as original payload. Then user would change `cwd` to `/home` and the resulting payload sent to the server would be `{cmd: 'ls -lah', cwd: '/'}`.

There is a problem though, the fact that besides 'current value' and 'previous value' there is also a default one (an empty `cwd` in case of running a new action) and with current implementation of re_run API endpoint, I see no way we can restore the original, empty value during rerun.

Kami: Yep, agreed - I was mostly just talking about the implementation (if we only send the changed parameters to the server, we should definitely display all of them either way).

In any case, sounds good.

About the default values - yes, user would need to explicitly specify (override) it - that's a feature :)

enykeev: I'm fail to see how it might be considered a feature, sorry =) In case of core.local cmd, the only way i can think of to restore the default value would be to inject something like `pwd`. And I'm still not entirely sure it would work.

Kami: re-run inherits all the parameters of the existing execution and if you want to override them you need to explicitly specify them - that's a feature.

If you want to use a default value, you need to explicitly override it like any other parameter. I'm definitely against using some kind of "magical" string to indicate a default value (e.g. null vs undefined, just no).

We do have a list runner types and list actions API so if you want to re-use a default value you need to call this first to retrieve a default value. I think that's perfectly fine and it's more explicit. And so far I haven't had a use case when I would want to re-run and action and use a default parameter value.

enykeev: So, any idea how to get a default `cwd` when rerunning `core.local` with `{cwd: '/'}`?

Kami: Actually, sorry, I'm slightly confused about this example:

> So, say we have an execution of core.local with parameters {cmd: 'ls -lah', cwd: '/'}. The rerun.payload would be the same as original payload. Then user would change cwd to /home and the resulting payload sent to the server would be {cmd: 'ls -lah', cwd: '/'}.

I assume here you meant "Then user would change cwd to /home and the resulting payload sent to the server would be {cmd: 'ls -lah', cwd: **'/home'**}."

As far as using default value goes - you would need to retrieve a default value from runner types API. Unless the default value is empty (None), in this case you are correct, you can't do that right now.

I can't think of a good approach to handle that - I'm against using a special string and I also don't like the approach of including additional attribute in the payload.

enykeev: > I can't think of a good approach to handle that - I'm against using a special string and I also don't like the approach of including additional attribute in the payload.

There's two, actually, though they are pretty close.

We either need to add `merge=false` query param to rerun API preventing it from merging payload parameters with previous execution parameters and then use it in st2client.js. We've already added so much special cases, another one makes absolutely no difference.

Or we should stop design our API around st2client.py needs and make `merge=false` the default behavior, then merge parameters client side in case of st2client.py (where you would need to retype them manually and thus this is necessary) or just pre-fill them in inputs in case of st2web.

Traditionally, there is a third option of ignoring this rant until the need in this functionality arise =)

Kami: Yeah, or in that case, you can just use "schedule and execution" API :P

enykeev:
If we have to plans to further extend rerun endpoint, then yeah, I guess I can. If on the other hand we would decide, for example, to add some metadata on the fact it is a rerun execution and the id of the original execution, then you would probably mind me doing it =)

apps/st2-history/controller.js
enykeev: If we have to plans to further extend rerun endpoint, then yeah, I guess I can. If on the other hand we would decide, for example, to add some metadata on the fact it is a rerun execution and the id of the original execution, then you would probably mind me doing it =)

Kami:
In any case, LGTM, +1. Let get that merged :)

If the default value limitation discussed above becomes an issue, we can always address it later.

PR #158, opened 2015-06-16
use run-local reporter for cloudslang flows

orius123:
@Kami this is the UI part for cloudslang integration. :smile:


Kami:
LGTM

/cc @enykeev

enykeev:
:+1:
Though eventually you may want to start showing child executions or add some kind of text representation of your workflow in action's code tab.

orius123:
agree. but in this phase of our integration it is not mature enough for supplying of all this data.
I suggest we will do it in next stages.

PR #164, opened 2015-07-01
Bump st2 version to 0.11.5

DoriftoShoes:


lakshmi-kannan:
+1

enykeev:
Please rebase over the latest master

manasdk:
We ty[ically avoid rebases over latest master. If we want some change from master better to cherry-pick. Likely that you know master will work with 0.11.5 in which case we can make an exception.

enykeev:
Honestly, I'm looking at the network graph and I'm lost. What's the point of us constantly updating version number in a separate branch if we aren't pulling new changes from master?

DoriftoShoes:
When we cut a minor release we branch st2web master since the assumption is that st2web master works with st2 master _at that point in time_.  When we release patches of st2 we need to bump the st2 version in package.json but we can't guarantee that the master st2web would work with last minor release.  In the past there have been breaking changes that would have blocked us from doing this (the streaming endpoint is an example).  I would much rather have st2web get "released" alongside the main platform releases than be constantly rebasing master back in to what has been deemed the "stable" branch.  We would undoubtedly end up breaking the web UI for the stable version.  If a bug fix needs to get in to the "stable" branch of st2web, it should be merged there as needed, not an entire rebase of master on a patch release of the platform.

PR #174, opened 2015-07-26
Fix rule url regression

enykeev:
STORM-1543 #resolve

### DIFF
--- a/apps/st2-rules/controller.js
+++ b/apps/st2-rules/controller.js
@@ -20,10 +20,10 @@ angular.module('main')
         url: '/new'
       })
       .state('rules.general', {
-        url: '/{ref:\\w+}/general?edit'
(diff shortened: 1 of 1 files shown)


apps/st2-rules/controller.js
```diff
@@ -20,10 +20,10 @@ angular.module('main')
         url: '/new'
       })
       .state('rules.general', {
-        url: '/{ref:\\w+}/general?edit'
+        url: '/{ref:[\\w.-]+}/general?edit'
```
Kami: Don't the dot and minus need to be escaped since we want to use it as a literal character?

enykeev: Square brackets ([character class](http://www.regular-expressions.info/charclass.html)) in regexp implies literal characters.

PR #181, opened 2015-08-20
Removing a limit on the execution workflow list

### LINKED ISSUE #180: Action/Details/History doesn't show full workflow
Only last few actions are shown, and NO indication given that there are more actions. 
The fix is to show ALL tasks in the workflow. 

This is what is shown: 
image

And this is the full history: 
image

emedvedev:
Fixes #180 — an execution now shows a full workflow when expanded from a list.

### DIFF
--- a/apps/st2-actions/controller.js
+++ b/apps/st2-actions/controller.js
@@ -219,7 +219,8 @@ angular.module('main')
       if ($filter('isExpandable')(record) && record._expanded) {
         st2api.client.executions.list({
           parent: record.id,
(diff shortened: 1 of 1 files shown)


enykeev:
:+1:
Check the latest changes to confirm whether it solves the latency problem you've reported and then merge at will.

emedvedev:
I think it works faster now, but I use workroom for API and not st2build, so maybe it just has less executions. The current issue is fixed anyway, and we can see if we still have latency problems later.

PR #182, opened 2015-08-21
[RFR] Showing icons in the action list

emedvedev:
Icons. Currently broken: can't figure out how to get data-url from image binary. Should work but it doesn't. Investigating. An alternative would be to just use the API call url as image source, but I'm not sure it's not going to raise problems with auth etc.

Note: Will only work with frozen pecan and the st2api PR I've pushed.

cc @enykeev

### DIFF
--- a/apps/st2-actions/controller.js
+++ b/apps/st2-actions/controller.js
@@ -60,20 +60,15 @@ angular.module('main')
           .value();
         _.forEach($scope.groups, function (value, key) {
           $scope.groups[key] = {  
-            'list': value
+            'list': value,
+            'icon': st2api.client.packFile.route(key+'/icon.png')
           };
-          st2api.client.packFile.get(key+'/icon.png').then(function(result) {
-            var blob = new Blob([result], {type: 'image/png'});
-            var reader = new FileReader();
-            reader.onloadend = function () {
-              console.log(reader.result);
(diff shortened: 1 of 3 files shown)


emedvedev:
Updated the PR.
It currently depends on:
1. Pecan being frozen at `stackstorm` fork.
2. st2 pull request https://github.com/StackStorm/st2/pull/1855
3. st2client.js pull request https://github.com/StackStorm/st2client.js/pull/32
Can be tested using the respective branches from the PRs.

cc @enykeev

apps/st2-actions/controller.js
```diff
@@ -58,6 +58,17 @@ angular.module('main')
           })
           .groupBy('pack')
           .value();
+        _.forEach($scope.groups, function (value, key) {
+          $scope.groups[key] = {  
+            'list': value,
+            'icon': st2api.client.packFile.route(key+'/icon.png')
+          };
+          st2api.client.packFile.get(key+'/icon.png').catch(function () {
```
enykeev: Instead of making two requests for every pack (one for an actual icon and one to check if it's there) you can just use `client.packs.list()` from StackStorm/st2client.js#31 to get the list of all the packs and associated files

emedvedev:
@enykeev That's way better. What about now?

emedvedev:
Merged into #186, closing.

Issue #176, opened 2015-08-05
Second level workflow indentation color shall alternate


dzimine:
Same color is used for indentation, making multi-level workflows difficult to parse visually. Suggest we alternate and use #9e53ea. Wish I knew CSS :) 

![image](https://cloud.githubusercontent.com/assets/1294734/9077566/85a25dce-3ae4-11e5-9458-57a3dedbefec.png)

PR #185, opened 2015-08-22
Hack to patch bug #176

dzimine:
Hack partially fixing  bug #176 

The right fix will take some Angular expertise. This may up to 3 levels of workflows.


enykeev:
Actually, this is exactly how I planned to do it all along. We can probably generate an extra class in Angular, but it would have no real benefit on that scale.

dzimine:
Than I go ahead merge it .

PR #186, opened 2015-08-22
Breaking rules by packs

emedvedev:
It would be best to merge this after the icons PR and not before.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@ node_modules
 components
 css
 build
+
+.htmlhintrc
--- a/apps/st2-actions/controller.js
+++ b/apps/st2-actions/controller.js
@@ -60,20 +60,15 @@ angular.module('main')
           .value();
         _.forEach($scope.groups, function (value, key) {
           $scope.groups[key] = {  
-            'list': value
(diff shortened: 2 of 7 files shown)


enykeev:
Ok, several problems:
- whitespaces. Line should not end on whitespace character which includes an empty line always having 0 length. Bonus points for editor config so we don't have to return to that ever again =)
- catch promises. Otherwise, the errors will go unnoticed.
- bump st2client.js to 0.3.6 so we would know this PR depends on the newer version of st2client and also requires user to make `npm install` again
- just a hint, `img:not([src])` is the selector you can probably make use of

enykeev:
this feedback is also related to #182

emedvedev:
Should be better now, fixed everything. I liked the `[src]` setup, but it's more readable with `:not`, you're right. :)
#182 can be left as is (or even closed), this PR merges that branch anyway.

enykeev:
LGTM, although, lets wait with merging until I publish 0.3.6. What to do with #182 is up to you to decide.

emedvedev:
Ok, I'll leave you to it. Merge whenever you're ready

PR #187, opened 2015-08-22
Make main menu respect "position" attribute in a state

emedvedev:
…which coincidentally puts history first in main menu. Just as planned!


enykeev:
:+1: though I still not particularly sold on the line of reasoning behind this order =)

emedvedev:
Well, at least now we have a mechanism to change it in five seconds if someone changes their mind again :)

Issue #183, opened 2015-08-21
Issue with rule edit


manasdk:
While creating the rule in the UI, we probably spotted some wrong behavior: Try to create a criteria based on the slack channel name (trigger.channel.name). While it is possible to do that during rule creation, the save button will be disabled if you ever want to edit the rule again and do not delete the criteria referring to trigger.channel.name again:

```
"criteria": {
    "trigger.text": {
      "pattern": "^Comment on <https://github\\.com/.*$",
      "type": "matchregex"
    },
    "trigger.channel.name": {
      "type": "contains",
      "pattern": "git"
    }
  },
 ```

Note : report came in via support

jonico:
The suggestion box in StackStorm does not show any nested attribute of trigger.channel, I guess this is related to the same root cause

emedvedev:
@jonico Same as #184—you can expect some progress later this week. :) Thanks a lot for reporting!

PR #191, opened 2015-08-25
"Namespace" module for trigger validation

emedvedev:
To tackle issue #183 I’ve added `st2-form-namespace` module which works similar to `st2-form-combobox`, accepts a list of suggestions and shows an autocomplete popup.

The key difference is that `combobox` only accepts exact match (for `trigger.channel` it will only accept `trigger.channel` in the field), while `namespace` will also accept everything within the object (like `trigger.channel.name`).

I feel like this approach is safer than just removing validation: we’ll retain the original schema while also allowing flexibility inside it.

cc @enykeev


emedvedev:
Closing this one, we'll use a different approach as discussed with @enykeev.

PR #192, opened 2015-08-25
Scroll to top after switching pages

emedvedev:
@manasdk: 
> 3) Hitting `next` on a list page loads new data fine but I remain at the bottom of the page. This is actually annoying as it requires me to to scroll to top of the page anyway - could we scroll to the top automatically?

Note: I don’t really like putting this in the main controller, so if someone (cc @enykeev) can point me to another place where it would be more appropriate, I’ll gladly change that.

### DIFF
--- a/apps/st2-actions/template.html
+++ b/apps/st2-actions/template.html
@@ -3,7 +3,7 @@
 <main class="st2-panel">
 
   <div class="st2-panel__view st2-actions" cg-busy="busy">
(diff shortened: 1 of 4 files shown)


emedvedev:
Added a todo for the future. @enykeev: looks good?

enykeev:
Looks awful, but yet again, we are already planning to fix that so yeah, +1

PR #195, opened 2015-08-26
Changing "enums" to "suggestions" and disabling validation in st2-form-combobox

emedvedev:
Another take on issue #183. Since we don’t use a strict schema, we don’t want to impose validation but should still have suggestions and autocomplete.

cc @enykeev

### DIFF
--- a/modules/st2-auto-form/modules/st2-form-combobox/directive.js
+++ b/modules/st2-auto-form/modules/st2-form-combobox/directive.js
@@ -87,7 +87,7 @@ angular.module('main')
 
   })
 
-  .directive('ngEnums', function enumDirective() {
+  .directive('ngSuggestions', function enumDirective() {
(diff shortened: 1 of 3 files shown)


enykeev:
LGTM. Please check why suggestions doesn't work in "Create rule" popup

emedvedev:
Fixed one more mention of "enums", suggestions work in "trigger" and "action" fields in the popup. Good to merge?

Note: suggestions don't work in "criteria" inside the popup for some reason, but this behavior is the same on `st2build001`. So, while it's not the scope of this PR, would still be nice to have it working.

enykeev:
I've rebased it against master to fix the problem with index.html formatting.

What I noticed though is that although we don't report any errors on criteria field having an error, the save button stays disabled.

PR #197, opened 2015-08-27
st2-view in all apps

### LINKED ISSUE #189: Make use of st2-view in actions and rules
As @manasdk reported, more often than not, we have a pretty long names for rules and it makes them truncate and became unreadable, especially on a lower resolutions. To fix that for sure, we would need to redesign the list view, but to ease the pain right now, the better option would be to port view menu to other apps besides executions.

screen shot 2015-08-25 at 10 10 44 am

emedvedev:
Closes #189


enykeev:
LGTM, but it is not "Cascade", but "Type"

emedvedev:
Changed to "type", merging.

PR #198, opened 2015-08-27
Showing UUID of execution

emedvedev:
We have to show execution ID and preferably one-click copy it.

Copying is not possible without Flash, but the alternative is putting UUID in a text input and selecting it on click. You’ll still have to press “Cmd + C”, but everything else is done for you. A solid alternative, I think.

### DIFF
--- a/apps/st2-history/template.html
+++ b/apps/st2-history/template.html
@@ -239,6 +239,14 @@
           <div class="st2-action-reporter__header">
             <span class="st2-label" status="record.status"></span>
             <div class="st2-action-reporter__timestats">
+              <div class="st2-action-reporter__execution">
+                Execution ID:
+                <input class="st2-action-reporter__uuid"
+                    type="text"
+                    value="{{ record.id }}"
+                    readonly="readonly"
+                    select-on-click />
+              </div>
               <div class="st2-action-reporter__timestats-started">
                 Started:
                 <time class="st2-action-reporter__time"
--- a/apps/st2-history/template.html
+++ b/apps/st2-history/template.html
@@ -234,39 +234,62 @@
         ng-class="{'st2-details__body--active': $root.isState(['^.general', '^.list'])}">
 
       <div class="st2-panel__scroller">
-
         <div class="st2-details__panel">
           <div class="st2-action-reporter__header">
-            <span class="st2-label" status="record.status"></span>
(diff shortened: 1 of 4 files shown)


emedvedev:
Also, had to merge action-output (#196) here, because otherwise there's really no place for the ID to go.

enykeev:
I don't really like the visual and I think something like that would look much better. 

Saying that, the rest LGTM.

enykeev:
Oh yeah, and I think you forgot to commit index.html. It generates just for convenience, the repository still should include the latest version.

emedvedev:
That's exactly why I didn't. Ok, I'll commit the latest.

Re visuals: feels a little bulky this way, doesn't it? I can use it, but I haven't seen it anywhere in the UI. Where's your screenshot from?

enykeev:
This is execution details from the execution triggered by the rule

emedvedev:
screenshot 2015-08-27 14 29 54
@dzimine @gmedrano @DoriftoShoes Want your feedback on this. Is the visual good? Is it easy to use? Aside from feedback on visuals, ready to merge.

Kami:
Aren't there some libraries which use new HTML 5 API for copying to clipboard and only fall back to Flash if that feature is not available?

In any case, it's not a big deal.

emedvedev:
http://zeroclipboard.org is the most common one and it's still flash 90% of the time. I wouldn't recommend it

emedvedev:
@manasdk gave a very vocal +1 in Slack, so I'm merging. :) Visuals can be improved later, after all.

dzimine:
@dzimine @gmedrano @DoriftoShoes Want your feedback on this. Is the visual good?
1) ways too sparse vertically, make them dense (1.0 interval) - else the rest of the info goes away from the view. 
2) rename `s` to `seconds`. 
Other than that - good.

manasdk:
> rename s to seconds. 
Meh. I like s/ms etc.

dzimine:
ok :) you get it. at least have a space `8 s` not `8s`

emedvedev:
@dzimine 
1. It's the same stylesheet as in execution triggered by the rule; should I make both more dense? If one's dense and another isn't, we lose consistency, that's not very good.
2. Added a space. :)

PR #201, opened 2015-08-28
Hotfix for rules editing

emedvedev:
Fixing rule editing.
cc @enykeev


enykeev:
LGTM. I myself would rather use first argument instead of `$scope.groups[group]`.

PR #203, opened 2015-08-28
Compute execution_time only if there is a finished timestamp

manasdk:
Done under the expert guidance of @enykeev


enykeev:
lol :+1: There's likely to be another problem related to the time not getting properly recalculated on execution completion, but that's minor and I'm hoping @emedvedev can take care of that Monday.

Kami:
Like a pro :P

PR #204, opened 2015-08-31
Showing pack name and icon in rule details

emedvedev:


Kami:
Screenshot would be nice.

In any case:

emedvedev:
buu6cual-zbckorjod-k_0vhwqkoc3ok1unchevs3w0

emedvedev:
Removed "pack" The reason why I didn't put it on the top right is that the rule name is actually supposed to be there, so we want to save that space. In the screenshot above it's not shown because the name is simply too long.

dzimine:
Ed, Gilberto is doing a full-redesign of the rule view, just be aware.

emedvedev:
Then I'll just go ahead and merge this, even if it's not very pretty.

dzimine:
yes.

PR #205, opened 2015-08-31
Putting "rule" subsection (along with enabled checkbox) on top of the edit view

emedvedev:
cc @dzimine 

I think it makes more sense in general to edit rule parameters first when editing a rule. And we put “enabled” checkbox on top that way.

### DIFF
--- a/apps/st2-rules/template.html
+++ b/apps/st2-rules/template.html
@@ -207,6 +207,21 @@
 
         <form name="form">
 
+          <div class="st2-details__panel"
+              ng-show="$root.state.params.edit">
+            <div class="st2-details__panel-heading">
+              <h2 class="st2-details__panel-title">Rule</h2>
+            </div>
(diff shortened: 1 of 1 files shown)


enykeev:
It was a conscious decision. User is here to wire a trigger to the action first and came up with the name at the end. You switch the two and you risk user forgetting why he does it in the first place.

emedvedev:
That's what I thought. @dzimine said that "enabled" checkbox should be visible "above the fold" though, so I wonder what we should do. I feel like duplicating the checkbox makes very little sense, and, to be honest, I'd still put "rule" section first because it's kinda unlikely somebody will forget the reason for editing something, but let's wait for more input :)

emedvedev:
@dzimine @gmedrano What do you say about this one? Should we have rule details first when editing?
I guess we'll redesign this anyway, but for now we should decide whether it makes more sense.

PR #206, opened 2015-09-03
Checking for "pack.files" before trying to show icons

emedvedev:


dzimine:
* show a default icon when no icon?
* longer-term: auto-generate an icon for pack if it's absent? This may be a backend work...

emedvedev:
We do show a default icon :) Some packs like "examples" don't have icons, so we use the one we've had in st2web before, just on a white background and a bit larger.

PR #207, opened 2015-09-03
[WIP] Redesigning rules to convey the "if/then" flow

emedvedev:
Need to work on styling—not very accurate at the moment—and there's no description for actions and trigger types (I'll probably need to change the API for that). The rest is done, and I think @gmedrano did a very good job. :)

screenshot 2015-09-03 16 47 34


Kami:
Cool :)

Where is the rule criteria and action parameters displayed? Still on the right or you can click on a small "box" on the left / right to display the criteria and action parameters?

emedvedev:
On the right. We've had a parameters line on one of the drafts, but it spanned the whole rule and @dzimine –1'd this "as [parameters] belong to action, and it's not clear visually". So it's going to be there eventually, we just need to rework the visuals :)

emedvedev:
screenshot 2015-09-09 17 50 18
@dzimine @gmedrano updated visuals and the API.

jfryman:
This is gonna be so killer. I'm excited!!!!

dzimine:
Come on Ed bring it in!!! :)

PR #208, opened 2015-09-03
Synchronize list of criteria operators

armab:
Add missing in WebUI `exists` and `nexists` from [st2 operators](https://github.com/StackStorm/st2/blob/f6e32185ead917e290f6f2489d45460fac5836df/st2common/st2common/operators.py#L187)

See: http://docs.stackstorm.com/latest/rules.html#critera-comparision

### DIFF
--- a/modules/st2-criteria/directive.js
+++ b/modules/st2-criteria/directive.js
@@ -39,7 +39,9 @@ angular.module('main')
             'td_lt': 'Earlier Than',
             'timediff_lt': 'Earlier Than',
             'td_gt': 'Later Than',
-            'timediff_gt': 'Later Than'
+            'timediff_gt': 'Later Than',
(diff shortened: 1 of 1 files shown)


Kami:
Good catch.

Wonder if we should just use operators metadata API (doesn't exist yet) to retrieve a list of available operators?

enykeev:
I'm saying it fom day one, we need an API to share settings\constants between client and server: jsonschemas, operators, supported features. We can go as far as sorting main menu apps or drawing forms entirely based on data provided by this endpoint so backend is free to change as much as you want and frontend is free to focus on making it work better rather than backporting backend changes.

emedvedev:
@enykeev are we merging this for now?
Also, why do we have two instances of "Exists" and "Not exists" with different operators? Just ran into that.

PR #209, opened 2015-09-04
Filter control

alex-shnayder:
Update `st2-filter` to work like filters in JIRA and GitHub. The control supports multiple choice, but it seems that API doesn't.

### DIFF
--- a/modules/st2-filter/directive.js
+++ b/modules/st2-filter/directive.js
@@ -1,31 +1,85 @@
 'use strict';
 
 angular.module('main')
-  .directive('st2Filter', function () {
+  .directive('st2Filter', function ($parse) {
 
     return {
       restrict: 'C',
       scope: {
-        type: '@',
-        label: '@',
-        filters: '='
+        defaultLabel: '@label',
+        multiple: '@',
+        items: '=',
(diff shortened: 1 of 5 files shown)


modules/st2-filter/directive.js
```diff
@@ -1,31 +1,85 @@
...
+          listElement.scrollTop = 0;
+          scope.search = '';
+          scope.refreshItemsToShow();
           element.toggleClass('st2-filter--active');
+          searchBarElement.style.minWidth = searchBarElement.clientWidth + 'px';
+          searchBarElement.focus();
         };
 
         scope.pick = function (name) {
-          var o = _.clone(scope.$root.state.params);
```
alex-shnayder: This stuff didn't belong here, so I moved it to the `updateFiltersAndGo` method in the root scope. Not sure if it's the best solution, though.

enykeev:
:+1:
Now it seems like we should limit min-width for `st2filter__list` to avoid it being to tight (ex. status). Or even just redesign them completely to make the whole thing lighter. Anyway, this change lifts the immediate pain of looking for the particular filter and that's good enough for now.

PR #212, opened 2015-09-08
Fix button appearance in Firefox

alex-shnayder:
1. Fix diagonal button borders not being antialiased in Firefox on OS X using a hack with `transform: scale` ([Bugzilla link](https://bugzilla.mozilla.org/show_bug.cgi?id=805393))
2. Fix off-by-pixel button height in Firefox by hardcoding `line-height`.


enykeev:
:+1:

PR #218, opened 2015-09-15
Fix rule Save button being disabled after removing criteria

alex-shnayder:


enykeev:
Oh shi... So, there were really no problem, we just forgot we disabled saving intentionally for when there were no changes has been made. lol. Not sure what would be the right thing to do, disable this "feature" or document it properly.

PR #221, opened 2015-09-16
UI refinements

emedvedev:
Lots of small refinements here and there from invision comments. Gifs will speak louder than words. :) (cc @dzimine @gmedrano)

oq7ln2
lfwr8m

Merging?


enykeev:
LGTM, though I'm pretty sure it would produce a bunch of conflicts when we rebase #228 on top of it.

PR #223, opened 2015-09-17
Error notifications

alex-shnayder:
Errors that were only displayed in the console now also show a notification popup.

### DIFF
--- a/apps/st2-actions/controller.js
+++ b/apps/st2-actions/controller.js
@@ -27,7 +27,7 @@ angular.module('main')
 
 angular.module('main')
 
-  .controller('st2ActionsCtrl', function ($scope, st2api, $filter) {
+  .controller('st2ActionsCtrl', function ($scope, st2api, $filter, Notification) {
 
     $scope.filter = '';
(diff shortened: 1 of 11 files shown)


enykeev:
I've spent unpleasant amount of time trying to figure out how to mock suitable responses with `mitmproxy` before I understood I simply wasting my time.

This thing works. It doesn't look especially in style with the rest. It is particularly annoying that the left edge blends with the edge of list view. It may have problems with plaintext responses (though we mostly eliminated them anyway).

It is probably ok to merge this thing and figure out how to properly test and style it later.

emedvedev:
@alex-shnayder can you do minor style adjustments? I'd appreciate it looking a bit more consistent, too. Everything else is good.

alex-shnayder:
Tried to improve the looks a bit, not sure if it's any better. Also, access control needs much more thinking.

/cc @enykeev @emedvedev

emedvedev:
Better, merging. We do need to provide better messages for RBAC scenarios, but that's a subject for another PR.

PR #228, opened 2015-09-22
Revamp rule details header

alex-shnayder:
screenshot-55

### DIFF
--- a/modules/st2-pack-icon/style.less
+++ b/modules/st2-pack-icon/style.less
@@ -1,4 +1,4 @@
-@size: 22px;
+@size: 32px;
 
 .st2-pack-icon {
   position: relative;
--- a/modules/st2-pack-icon/style.less
+++ b/modules/st2-pack-icon/style.less
@@ -1,0 +1,44 @@
+@size: 22px;
+
+.st2-pack-icon {
+  position: relative;
+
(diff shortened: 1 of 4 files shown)


dzimine:
@alex-shnayder pls take a look https://stackstorm.slack.com/archives/ui-ux/p1442951276000941

alex-shnayder:
@dzimine Can't see it, probably no access.

enykeev:
The minor polish would be to compensate the paddings in proportion 16px 20px 16px (http://cloud.mech.sh/dJEq) rather than current 20px 20px 11px (http://cloud.mech.sh/dK4Z), but the rest seems fine by me.

modules/st2-pack-icon/style.less
```diff
@@ -0,0 +1,44 @@
+@size: 22px;
```
enykeev: Hey, wait a second, who said it is 22? Nope. Despite @gmedrano artistic vision, 32px icons are non negotiable.

enykeev:
diff

emedvedev:
Smaller icons look better, but I side with @enykeev on this one: we can't expect browser scaling to look good every time, so we have to resort to natural size which is 32x32. That said, everything else looks good.

PR #229, opened 2015-09-22
Fix execution parameters clipping

alex-shnayder:
In the History list, execution parameters now don't get cut if there is enough space.

### DIFF
--- a/apps/st2-history/controller.js
+++ b/apps/st2-history/controller.js
@@ -305,14 +305,11 @@ angular.module('main')
   .filter('fmtParam', function () {
     var fmtParam = function (value) {
       if (_.isString(value)) {
-        return '"' + (value.length < 20 ? value : value.substr(0, 20) + '...') + '"';
+        return '"' + value + '"';
       }
 
       if (_.isArray(value)) {
-        return '[' +
-        _(value).first(3).map(fmtParam).join(', ') +
-        (value.length > 3 ? ',..' : '') +
-        ']';
+        return '[' + _(value).map(fmtParam).join(', ') + ']';
       }
 
       return value;
--- a/apps/st2-history/style.less
+++ b/apps/st2-history/style.less
@@ -74,6 +74,14 @@
     }
   }
 
+  &__column-action-param {
+    display: inline-block;
+    overflow: hidden;
+
+    white-space: nowrap;
+    text-overflow: ellipsis;
+  }
+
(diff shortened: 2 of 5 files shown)


enykeev:
There's some minor styling issues: I myself prefer not to align indentation in multi-line statements because it would result in lots of noise on variable name change and it sometimes require you to add an odd number of spaces. Instead, you should split long statements into several short ones if possible or just add another level of indent when shifting them to the next line. As I said, minor.

Another thing is that I myself trying to avoid things like `containerStyle.getPropertyValue` and `container[0].clientWidth` because there's no guarantee they will behave the same in every browser. Just the other day we spent a ton of time trying to figure out why we getting float values from our layout function in some browsers and not the other. Moreover, taking into consideration the whole complexity of `makeProportional()`, I have to confidence it will not stuck in while loop forever in some cases.

But most importantly, I don't see whether there's a real difference between this solution and just using flexbox http://codepen.io/anon/pen/dYpoPb.

I mean, yeah, I get it. We've asked for more sophisticated algorithm for calculating parameters length and we got it. It does its job and nothing is perfect. So I wonder if there's actually a problem or am I just picky? And is this a time to be picky? @dzimine @emedvedev

alex-shnayder:
There is quite an important difference between flexbox and my solution: 

parameters-clipping

Flexbox clips all children elements at the same time, while my algorithm starts clipping the widest element until it reaches the width of the second widest. Then both of them are clipped, then three, etc. I don't have much experience with flexbox though, so I might be missing something. 

That being said, I agree that it might be an overkill. Maybe a better solution would be to redesign the element rather than to try to squeeze everything into the current shape.

enykeev:
I see. Yeah, you're right, I don't know how to give each of them at least an even portion... It is likely possible with some minimal JS hackery, but nah...

The idea to redesign this thing also crossed my mind the moment I've seen what Gilberto did with rules. We'd better keep executions low profile since there might be a ton of them (about half a million on stage1 if memory serves me well, plenty to get tired of scrolling), but we can at least try to apply the same idea and make the parameters field full width and scrollable.

We'll get back to it next week...

emedvedev:
I think it's fine: not ideal, but gets the job done, and current behavior doesn't look good in some cases, so I'll go ahead and merge this, we can refine later (or, as proposed, redesign it completely when the time comes).

PR #230, opened 2015-09-22
Show rule definitions as YAML

alex-shnayder:
Convert rule definitions to YAML on the fly. This way _all_ rules will be displayed as YAML, but it won't be exactly the same code as in the file. The performance is good, the color scheme needs adjusting.

cc @enykeev @dzimine

### DIFF
--- a/apps/st2-rules/template.html
+++ b/apps/st2-rules/template.html
@@ -287,7 +287,7 @@
         ng-class="{'st2-details__body--active': $root.isState('^.code')}">
       <div class="st2-panel__scroller">
 
-        <div class="st2-highlight" code="rule | json" lines="20" language="javascript"></div>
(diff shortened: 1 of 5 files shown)


dzimine:
Suggest we hold this for 2 days, to get over Launch AMI, and fix after.

emedvedev:
@enykeev @dzimine are we merging this? It's questionable: we're not taking YAML directly from the file, we're converting JSON, so it can lead to a certain degree of confusion. I don't have a strong opinion on this, but I remember objections to this approach.

emedvedev:
@alex-shnayder we discussed JSON/YAML and will stick to the solution in this PR (displaying everything as YAML), so I'm ready to merge once you adjust the color scheme. Do you think you can do this on your own or do you need @gmedrano's help with the colors/visuals?

emedvedev:
Merging.

PR #231, opened 2015-09-23
Make all routing calls remember filters

enykeev:
The rule of thumb is that unless you specifically want to override the filters, you should never use `$root.state.go`. Consider it low level API.


emedvedev:
Just stumbled upon this PR.
Note taken. Also, if I'm in a position to review, LGTM. :)

PR #232, opened 2015-09-29
Make rule edits apply instantly on the whole page after save

alex-shnayder:
There is one issue. When editing a rule, its json representation is updated in real time before the rule is even saved. But some fields like _uid_ and _trigger.ref_ are only updated after saving. Not sure what's the best way to fix this.

### DIFF
--- a/apps/st2-rules/controller.js
+++ b/apps/st2-rules/controller.js
@@ -243,13 +243,14 @@ angular.module('main')
 
     $scope.submit = function () {
       st2api.client.rules.edit(angular.copy($scope.rule)).then(function (rule) {
+        $scope.rule = rule;
         $scope.form.$setPristine();
         $scope.form.saved = true;
 
         _($scope.groups).forEach(function(n, group) {
-          var index = _.findIndex($scope.groups[group], {'ref': rule.ref});
+          var index = _.findIndex($scope.groups[group].list, {'id': rule.id});
           if (index >= 0) {
-            $scope.groups[group][index] = rule;
+            $scope.groups[group].list[index] = rule;
             $scope.ruleMeta = _.clone(rule);
           }
         });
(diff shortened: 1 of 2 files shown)


emedvedev:
@enykeev thoughts?

enykeev:
LGTM as long as it works, though I wonder whether we should have instead got rid of https://github.com/StackStorm/st2web/blob/rule-instant-update/apps/st2-rules/controller.js#L135-L137

emedvedev:
@alex-shnayder We do need to update triggers: while it's the property that probably gets changed the least, it's still strange and unexpected behavior, because a trigger is there but its description is not. So maybe capturing a "post-save" event, or fetching the rule again after save would be a way to solve this? @enykeev and @manasdk can probably contribute to the discussion.

alex-shnayder:
@emedvedev I'm not sure I understand you (or you understood me). Most of the changes to a rule are reflected in its JSON definition instantly after you change the value of a field. But there are some properties that cannot be changed directly in the interface, they are generated server-side. These properties only get updated in the JSON definition after you click the Save button. 

Let's say, for example, that you have a rule with _examples.event_ trigger. Its JSON definition looks like this:

```JSON
"trigger": {
  "ref": "examples.event",
  "type": "examples.event"
}
``` 

Then you switch the trigger to _core.st2.IntervalTimer_. At this point JSON becomes inconsistent:

```JSON
"trigger": {
  "ref": "examples.event",
  "type": "core.st2.IntervalTimer"
}
```

After you click Save, a request is made and the definition is updated to the consistent state:

```JSON
"trigger": {
  "ref": "core.st2.IntervalTimer",
  "type": "core.st2.IntervalTimer"
}
```

Everywhere else on the page changes are only applied after saving.

emedvedev:
I meant a slightly different issue — when you update a trigger, its description does not get updated in the rule list, it's just empty. That needs to be fixed, but refs and uids are less important, we can merge without updating them.

alex-shnayder:
Fixed.

alex-shnayder:
Rebased to fix an issue in newer code.

emedvedev:
+1.

PR #237, opened 2015-10-07
Fix combo boxes not working in the new rule form

alex-shnayder:
`Ng-if` creates an isolated scope, so `$parent.` prefix is needed when using `ng-model`.


emedvedev:
LGTM

PR #238, opened 2015-10-12
Pack field in rule forms

alex-shnayder:
The way `st2-form-combobox` is used here is kinda hacky since it cannot be used in an auto form. I think it would be better to make all the inputs independent from `st2-auto-form`, but it's a whole 'nother story. 
Also, combo boxes now allow any values even if they are not in the suggestions. An option for restricting this would be useful.


emedvedev:
LGTM.

PR #241, opened 2015-10-16
Fix safari variable shadow error

enykeev:
For some undocumented reason, safari don't like expressions like `function e(e) {}` up to the point it considers them an error. Minified code of one of our dependencies had this code fragment which prevented the wole codebase from being parsed in safari.


enykeev:
@alex-shnayder pay attention, you don't see things like that every day =) Thankfully.

jfryman:
Good find.

PR #242, opened 2015-10-19
Support type object field

### LINKED ISSUE #233: WebUI doesn't present parameters for action input when they are objects
How to reproduce:
1. Create an action with a python dictionary parameter.
2. Declare parameter type as object in the action yaml file.
3. Load action and try to run it in UI.

enykeev:
Resolves #233

Added special type for `type:object` based on auto-resizing textarea module. The field expects JSON string as input and promptly shows an error when there's a syntax error.


emedvedev:
LGTM, although I'd ask more people to test if we have time.

PR #246, opened 2015-10-28
Add Zombie tests for every application

enykeev:


enykeev:
I've updated the tests, but they still require an st2 server to run against. For now, you can manually run them by defining a number of ENV variables:
```
ST2_HOST=192.168.10.138 ST2_USERNAME=admin ST2_PASSWORD=123 npm test
```

PR #259, opened 2015-11-27
Overhaul flex tables collapsing

alex-shnayder:
@enykeev please, review.


enykeev:
Looks good to me. The only thing we need to be cautious about with services is to avoid passing DOM references to it as it will result in massive memory leaks. Doesn't look like the case, though.

PR #261, opened 2015-12-03
adding re:amaze KB widget to st2web

jfryman:
This PR adds the necessary bits to get the re:Amaze KB tool loaded and ready to go during day-to-day usage.

/cc @enykeev - keep me honest, here. :)


enykeev:
Don't see it in UI. And you also didn't remove our old "report bug" button. Will take a closer look after I'm done with the rest.

emedvedev:
Works well. Merging.

Would be nice if we could default the KB search to the "web" section automatically, maybe. Not sure.

enykeev:
no, it won't. We've a number of issues in this repo that doesn't belong to st2web rather st2 or some other component and only end up here because we default to this repo when reporting an error in UI.

emedvedev:
I was only referring to the Knowledge Base search, not sending a support request, but I see your point. :)

jfryman:
:bow:

PR #269, opened 2016-01-12
CircleCI integration

enykeev:


circle.yml
```diff
@@ -0,0 +1,62 @@
...
+deployment:
+  publish:
+    owner: StackStorm
+    branch:
+      - master
+      - /v[0-9]+(\.[0-9]+)*/
+      - feature/circleci
+    commands:
+      - dpkg-buildpackage -b -uc -us
+      - rpmbuild -bb rpm/st2web.spec
```
armab: I think these 2 lines should be part of `test` section.

This will allow to check if packages are able to build if you/or someone open PR against these `master` or `vX.Y.Z` branches.

circle.yml
```diff
@@ -0,0 +1,62 @@
...
+# BINTRAY_ORGANIZATION
+# BINTRAY_ACCOUNT
+# BINTRAY_API_KEY
+general:
+  # Don't run CI for PR, only for major branches
+  branches:
+    only:
+      - master
+      - /v[0-9]+(\.[0-9]+)*/
+      - feature/circleci
```
armab: Same as previous note.

I think this limitation on build could be removed, so we run tests on every PR, since it's really fast for `st2web`.

I assume this `circle.yml` file is from `st2` branch (now it's still expensive to run build on every PR).
For `st2-packages` we run builds on every PR: https://github.com/StackStorm/st2-packages/blob/master/circle.yml#L8

armab:
I've added under https://circleci.com/gh/StackStorm/st2web/edit#env-vars ENV Variables related to Bintray, which are required by `bintray.sh` to upload packages.

But I'd suggest to create your account in https://bintray.com/ so I can invite you/give access in organization there. I guess it's interesting to see everything from A to Z)

circle.yml
enykeev: makes sense, yeah, but I guess we should still keep `deployment.branch.only` since we likely don't want a new package on bintray for every PR.

circle.yml
enykeev: so do I understand correctly that `exit 1` during deployment phase would still fail the run, but what you're saying is that this run won't even start for branches except for `general.branches.only`?

armab:
I can bring commits for deployment if you want tomorrow only, but it's nothing really hard here. If you want to go through entire thing to understand it better - I'm for it :+1: and can guide you.

Regarding `circle.yml`, what we'll need: 
1.) Deploy single `.deb` file to `trusty_staging`, `wheezy_staging`, `jessie_staging` Bintray repos
2.) Deploy single `.rpm` file to `el6_staging`, `el7_staging` Bintray repos.

Looks like duplication, but it's useful to install all packages from 1 repo, without adding second one during installation.

Nice to copy `.deb` and `.rpm` files to `~/packages` artifacts dir. So we'll have artifacts in CircleCI too: https://circleci.com/docs/build-artifacts (just nice feature to have)

Also we'll need to discover `st2web` revision number for the next build, see: https://github.com/StackStorm/st2-packages/blob/master/.circle/buildenv.sh#L47, seems `bintray.sh` needs adjustments here. I'll bring the change tomorrow, but for now `1` revision should work well (not a blocker).

----

If you want to switch to other task like flow - I'm more than OK to finish Bintray/Circle tomorrow. So only depends on your interest/curiousity here :cake:

circle.yml
armab: Right

enykeev:
Yeah, I guess I'm actually interested in going all the way through. I wonder though at what point staging packages becoming production ones? Do we have a tooling for that, are we planning to do it in the future or are we going to do it manually?

Also, do you mind pointing me where exactly are we using payload.json?

circle.yml
armab: Correct.

If we keep `build package` logic in deployment section there could be such scenario: 

- Open PR (`test` section passed -> `deployment` is not executed -> green)
- Merge PR into `master` (`test` passed -> `deployment` executed and failed -> red). Need to revert changes, fill another PR that will fix `rpmbuild` or `dpkg-buildpackage` logic.

By that reason it's better to move package creation logic into `tests` section and leave for `deployment` only simple uploading.

armab:
Yeah, so in this picture: 

https://github.com/StackStorm/discussions/issues/111

CircleCI work is to build and deploy packages to Bintray Staging repos.

Once they're in Staging the rest work is up to `st2` box to install those newly created packages, run all important tests like acceptance/end2end and only after that if everything OK -> deploy tested packages to Bintray Production and Docker Hub.
So the `st2` staging box should decide if packages are good enough for production or not.
See execution plan here, Staging CI: https://github.com/StackStorm/discussions/issues/134 is up to Lakshmi and Manas.


The `payload.json` is the file which contains more detailed metadata of CircleCI build, needed for `st2` staging/build box to consume. The important is information `which` version and revision number of packages were just created.
For example: https://circleci.com/gh/StackStorm/st2-packages/199
Find respective section: 
In fact `payload.json` is a [CircleCI artifact](https://circleci.com/docs/build-artifacts), ex: https://circle-artifacts.com/gh/StackStorm/st2-packages/199/artifacts/0/home/ubuntu/packages/payload.json

circle.yml
```diff
@@ -0,0 +1,62 @@
...
+# BINTRAY_ORGANIZATION
+# BINTRAY_ACCOUNT
+# BINTRAY_API_KEY
+general:
+  artifacts:
+    - ~/packages
+
+machine:
+  environment:
+    DISTROS: "wheezy jessie trusty el7"
```
armab: Add also `el6`, we have that repo.

In next few days we'll start shipping RHEL6/CentOS6 packages in https://bintray.com/stackstorm/el6_staging

PR #277, opened 2016-01-22
remove circle_ci

manasdk:


manasdk:
it is ok to leave circle.yml on this branch. We plan to ship packages from here :)

PR #278, opened 2016-01-27
Change production defaults

enykeev:


config.js
```diff
@@ -2,13 +2,4 @@
 angular.module('main')
   .constant('st2Config', {
 
-    hosts: [{
```
dzimine: hm.. may be comment them out? Users need hints. Discoverability is a problem when we have "nothing".

enykeev: Makes sense, I guess. Just to be on the safe side.

PR #281, opened 2016-02-03
Move product defaults to 1.3

dzimine:
To make 1.3 package work.

PS. I should propagate this "nicer" example to master.


estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1454544446004953

PR #283, opened 2016-02-11
Get rid of bower

enykeev:
Seems like every package we need is now available in npm. Time to bid our friend Bower farewell. Never liked it very much.

Keeping bower.json for the time being so we don't break our build pipelines. An absence of `dependencies` makes `bower install` a no-op.


emedvedev:
Tested on a fresh install, seems to be working. LGTM. I'll play with it some more and merge by the end of the day.

enykeev:
We would also need to check that it will continue to build everywhere.

emedvedev:
Tested and made a dry-run build. Merging.

Kami:
:wave:

PR #285, opened 2016-02-15
Enable deployments to PackageCloud (1.3)

armab:
The intent of this PR is to enable deployments to PackageCloud for `stable` `1.3` version.

See https://github.com/StackStorm/st2web/pull/284
(cherry picked from commit 8b2d22e)

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -44,10 +44,12 @@ checkout:
 
 dependencies:
   pre:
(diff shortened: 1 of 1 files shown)


armab:
Build failed, so I probably won't use cherry-picked commit here, but will work based on `v1.3` `circle.yml`.

PR #287, opened 2016-02-24
Remove bintray integration

enykeev:


armab:
Personally I'm +1 for this in general, and to do the change like that in all repos, but you need another +1 from @lakshmi-kannan 


As he noted, there should be some transition time. It's just too early to completely switch, because in docs we still don't reflect new PackageCloud repos and there are users who have Bintray packages installed.

lakshmi-kannan:
I am +1 as I fixed the docs and bootstrap script to package cloud.
See https://github.com/StackStorm/st2docs/pull/63#issuecomment-188440871 and https://github.com/StackStorm/st2docs/pull/63#issuecomment-188441491

lakshmi-kannan:
Can we fix other repos as well to not publish to bintray?

armab:
Sure, I'll prepare tomorrow PRs for `st2-packages`, `st2` branch `master` and `st2` branch `v1.3.3`

PR #306, opened 2016-05-17
Exclude unnecessary attributes from Action List request

enykeev:
A companion change to StackStorm/st2#2685


Kami:
Building failing, besides that, LGTM, :+1:

PR #308, opened 2016-05-18
Properly serialize default value to state value string

enykeev:
Failure to do so results in default value of `{}` being represented in a field as `[object object]`.


Kami:
:+1: 

(although, without tests, did it really happen? :P)

enykeev:
lol you are persistent aren't you =)

yeah, you're right, it's time to stop slacking on tests

PR #309, opened 2016-05-18
Expand pack when user tries to access the action directly

enykeev:
By default, we show all the packs collapsed to help user better navigate through the vast amounts of actions modern st2 installation may have.

Although, in case user lands on the page with the URL pointing to the specific action (like in http://192.168.10.26:3000/#/actions/packs.load/general), he likely wants to see this action on the screen and selected.

This PR ensures the pack for currently selected action gets expanded.


Kami:
LGTM

Issue #310, opened 2016-05-18
Other color for cancel button


johandahlberg:
Canceled workflows are displayed in a grey color very similar to the oscillating grey for running workflows. To make the two easier to distinguish between I'd suggest displaying it in a different color. Perhaps yellow to sync with what is displayed in the CLI? I can see if this is not high up on the priority list, but I just thought I'd mention it.

Keep up the good work!

enykeev:
Resolved in https://github.com/StackStorm/st2web/pull/316

johandahlberg:
Awesome!

PR #316, opened 2016-07-05
Add a label for canceled executions

enykeev:
screen shot 2016-07-05 at 2 58 32 pm


enykeev:
Resolves https://github.com/StackStorm/st2web/issues/310

PR #322, opened 2016-08-12
Restyle with respect to Brocade style guide

enykeev:
screen shot 2016-08-23 at 12 20 36 pm
screen shot 2016-08-23 at 12 20 52 pm
screen shot 2016-08-23 at 12 21 04 pm


Kami:
Nice! :+1:

LindsayHill:
Good work. One minor thing: I think we'll need to change the colour for the highlighted tab (History/Actions/Rules) in the header. Should be something in the Brocade palette that complements the main colour.

LindsayHill:
Overall I think the third one (red header) probably makes it feel most like a "Brocade" app.

enykeev:
@LindsayHill DZ have picked red one for bwc-ui and blue one for st2web

PR #324, opened 2016-08-25
Compact view for history records

enykeev:
screen shot 2016-08-25 at 7 58 19 pm

### DIFF
--- a/apps/st2-actions/style.less
+++ b/apps/st2-actions/style.less
@@ -63,6 +63,17 @@
   &__executions {
     .st2-flex-table {
       margin-bottom: 10px;
+
+      border: 0;
+    }
+
+    .st2-history-child {
+      border: solid @grey-lighten-3;
+      border-width: 0 1px 1px 0;
+    }
+
+    .st2-history-child .st2-history-child {
+      border: 0;
     }
   }
 }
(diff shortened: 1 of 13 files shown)


enykeev:
I don't like that at all, to be honest. This zigzag line now looks out of place, the notion of if-then is lost and the logic in which the line breaks in two when it gets too long makes absolutely no sense comparing to Rules.

I'm not even talking how I had to add a number of styles so everything else would not just fall apart.

I'm putting it here you you can judge yourself. My call, lets get to it in 2.1.

armab:
True, looks pretty strange.

lakshmi-kannan:
enykeev: I agree with you. Let's target this for 2.1. However, for me, this is better than timestamp in new line :P.

dzimine:
This is unacceptable.

let's talk tonight, wire Davide if we can, brainstorm how to fix it, 
and iterate till we like it.

enykeev:
Seems like after some discussion, we finally managed to find a compromise that would work. We're basically rolling back to the same order of columns we had before the repaint.

screen shot 2016-08-26 at 2 17 47 pm
screen shot 2016-08-26 at 2 18 20 pm

I've also put few more changes (based on @dzimine feedback) to that PR.

enykeev:
I'm gonna merge that to make sure all our packages are building and installing correctly. If you have more feedback, leave it there, and we'll get back to it as soon as we have time.

lakshmi-kannan:
@enykeev I like the compromise. It's looking good. I definitely missed it. These changes are against st2web. Does that mean even community people get to see Brocade colors? 

I didn't see these changes being picked up for bwc-ui. So I am a little confused on what the plan is.

PR #325, opened 2016-08-26
Explicit limit and offset for execution calls

enykeev:
Due to changes in api validation, we're now required to clearly state limit and offset.


enykeev:
Actually, I don't know when we made this change and there's surely better ways to get around it, but at that point, I just want it gone.

PR #326, opened 2016-08-29
Make workflow icon bigger

enykeev:
screen shot 2016-08-29 at 6 07 18 pm


dzimine:
Merging as is, will figure better in 2.1

dzimine:
Spoke with @emedvedev and agreed with him saying "It's OK Dmitri" :) But we need to call it out in 2.1 design review.

PR #327, opened 2016-09-09
Update the form on entity change

enykeev:
Fixes https://github.com/StackStorm/st2/issues/2891.

The problem is related to how react uses virtual DOM to determine which DOM nodes should be updated. Before the change, the only indicator of whether the field should be updated was its name so if we're to switch between two entities having a field with the same name, the element representing the field would not be updated.

In most cases, the field value could still end up being updated due to change in `$scope.payload` object (we set it to `{}` on entity change), but there's two problems with that:
 - one of our dependencies, `react-textarea-autosize` doesn't support this kind of update mechanics and would require us to push the change upstream
 - while we're in transition between angular and react, we're using `ngreact` module to piece them together. The module requires us to define `watch-depth` on the element separating the two. For field value to update, we would need to change `watch-depth` from `reference` to `value` for changes to propagate and it is not clear how it would affect some of our custom fields.

For now, I've decided to set another `key` for the whole form itself which would force it to be redrawn on every entity change. This isn't the prettiest solution, but at the end, this is what we expected to happen all this time anyway.

### DIFF
--- a/apps/st2-actions/template.html
+++ b/apps/st2-actions/template.html
@@ -156,6 +156,7 @@
                 Action has no mutable parameters
               </div>
               <div class="st2-auto-form"
+                key="action.ref"
                 data-test="action_parameters"
                 watch-depth="reference"
                 on-change="onChange"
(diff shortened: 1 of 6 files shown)


lakshmi-kannan:
TBH, I don't get it because I am a noob wrt Angular and React. But you have my +1 to merge if it addresses the problem.

PR #341, opened 2017-01-05
Update rule criteria operators

Kami:
The definition was missing new `regex`, `iregex` and `matchwildcard` operators. In addition to that, it still contained deprecated `matchregex` operator.

In the future, we should expose meta / view API in st2 to retrieve available operators so we don't need to hard-code it here (perhaps `GET /v1/rules/views/operators` or similar).

Caught and reported by Lisa Bekdache on Slack.

### DIFF
--- a/modules/st2-criteria/criteria.directive.js
+++ b/modules/st2-criteria/criteria.directive.js
@@ -22,7 +22,9 @@ module.exports =
 
         scope.typeSpec = {
           enum: {
-            'matchregex': 'Matches Regex',
+            'regex': 'Regular expression match',
+            'iregex': 'Case-insensitive regular expression match',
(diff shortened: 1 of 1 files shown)


enykeev:
We absolutely should not create multiple endpoints exposing different specifics of particular st2 installation. There should be a single endpoint, something like `/v1/settings` that would tell everything that client might need to know: st2 version, operators, auth backends, things like `mask_secret`, `max_page_size`, `debug`, list of available runners, maybe even their versions, presence of mistral, stream, st2auth. If we are to gather all this information from different endpoints, it might take us dozens of requests to do so and some of it might be critical to the render further decreasing UI responsiveness.

PR #344, opened 2017-01-23
Update runner types in action reporter

enykeev:
The renaming done for 2.0 release has not been properly propagated to st2web.


bigmstone:
Now your question makes more sense.

PR #349, opened 2017-02-10
Update Brocade fonts

emedvedev:
This PR adds an empty `apps/extensions.js` file that can be overwritten to include additional apps in st2web.


emedvedev:
https://github.com/StackStorm/discussions/issues/223

enykeev:
I say that you don't even need `extensions.js`, you can just override `apps/index.js` file. The applications are mostly self-contained. If you as a suite developer decide that user should not have an access to some default applications, you need a way to disable them too.

From my point of view, suite developer should be the one deciding what should be delivered to his users. He surely know them better than we do. At the same time, he should also be made responsible for compatibility between new version of st2web and the suite. We can't do it for him.

emedvedev:
@enykeev the problem I couldn't solve (that will still stand with overriding `apps/index.js`) is the inclusion of several suites, since suites would be developed independently, but can be installed together: that would be two packages overriding the same file, and the suite you install last would "win". :) `extensions.js` is just easier to change than `apps/index.js` when it becomes a problem.

Ideally, we would do something like `require('st2-*').name`, but wildcard requires aren't supported for obvious reasons.

enykeev:
That's exactly the problem I want us to make obvious by editing `apps/index.js`. "For now, there is no magic so you have to cooperate". `Extensions.js` has the same problem, but it's more vague.

emedvedev:
:+1: I'll get back to using `index.js` then.

emedvedev:
@enykeev removed the extensions file, so now all this PR does is updates Brocade fonts. :) I'll modify the one in bwc-ui as well.

dzimine:
👍

Issue #355, opened 2017-05-11
Text in select boxes are not visible until mouseover


pietervogelaar:
On Windows, text in select boxes are not visible until mouseover.  This is both on Firefox and Chrome. In Internet Explorer only a white page is visible. Is Internet Explorer not supported at all for st2web?

On Mac everything works great with Chrome and Firefox.

estee-tew:
https://stackstorm.slack.com/archives/community/p1494879221234565

LindsayHill:
Confirmed: OS-specific. Fails when Windows 10 used, works with macOS.

LindsayHill:
Also, seems to have been problem since at least st2 2.1.1 (my test was with BWC 2.1.1 system)

nzlosh:
I have the same issue under Linux with FF 45.9.0 ESR http://imgur.com/a/4wxfo
Looking at the style information, this where it's getting set to white
```
.st2-auto-form__field option {
    color: white;
}
```
If I set `color` to black the drop down box shows all entries.

LindsayHill:
Fixed by #356

bigmstone:
@pietervogelaar @nzlosh This was fixed in #356. Let us know if this does not resolve the issue.

PR #356, opened 2017-05-15
Make criteria selection visible on Windows

LindsayHill:
Drop-downs don't seem to be visible on Windows - see issue #355. It works on macOS, but not Windows

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -10,7 +10,7 @@ machine:
     DEPLOY_PACKAGES: 1
     DEB: wheezy jessie trusty xenial
     RPM: el6 el7
-    ST2_TEST_ENVIRONEMENT: https://github.com/enykeev/st2box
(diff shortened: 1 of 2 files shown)


LindsayHill:
Updating npm seems to work around NPM issues, but now it looks like the nginx container doesn't have openssl installed by default any more
> /bin/sh: 1: openssl: not found
>ERROR: Service 'web' failed to build: The command '/bin/sh -c mkdir -p /etc/ssl/st2 &&   openssl req -x509 -newkey rsa:2048 -keyout /etc/ssl/st2/st2.key -out /etc/ssl/st2/st2.crt -days 999 -nodes -subj "/C=US/ST=California/L=Palo Alto/O=StackStorm/OU=Information Technology/CN=$(hostname)"' returned a non-zero code: 127

LindsayHill:
nginx container now pinned to 1.11, so these tests pass.

PR #360, opened 2017-06-13
Handle Array of Objects

### LINKED ISSUE #352: st2web doesn't handle array of objects parameters correctly
For details, see https://github.com/StackStorm/st2/issues/3362.

bigmstone:
Fixes #352 

Note: This does change st2web workflow to require `[` `]` wrappers for list.

### DIFF
--- a/modules/st2-auto-form/tests/test-array.js
+++ b/modules/st2-auto-form/tests/test-array.js
@@ -30,11 +30,11 @@ describe('AutoForm ArrayField', () => {
 
     const c = new TestComponent(<ArrayField {...props} />);
 
-    c.makeChange('t,e,s,t');
+    c.makeChange('["t","e","s","t"]');
 
     expect(onChange.withArgs(['t','e','s','t'])).to.be.calledOnce;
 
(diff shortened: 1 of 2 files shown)


modules/st2-auto-form/tests/test-array.js
```diff
@@ -30,11 +30,11 @@ describe('AutoForm ArrayField', () => {
 
     const c = new TestComponent(<ArrayField {...props} />);
 
-    c.makeChange('t,e,s,t');
+    c.makeChange('["t","e","s","t"]');
```
Kami: Do we still handle case without brackes in the same manner? Is this even desired now?

bigmstone: No. It needs brackets now as indicated by the description. I could have handled those cases by first attempting to JSON parse the output then on failure attempting to split on `,` but this would be far less explicit and error prone. Additionally the UX for `object` requires braces to be included so this feels more consistent in my opinion.

dzimine: now (after much discussion) we should: 
https://github.com/StackStorm/st2web/issues/394

PR #363, opened 2017-06-13
Sorting params by required then by alphabetical order

### LINKED ISSUE #154: Parameters ordering
When showing list of parameters for specific action via CLI, the parameters are ordered first by `alphabet`, and then by `required`: 

``` sh
$ st2 run core.http -h

Action that performs an http request.

Required Parameters:
    url
        URL to the HTTP endpoint.
        Type: string

Optional Parameters:
    allow_redirects
        Set to True if POST/PUT/DELETE redirect following is allowed.
        Type: boolean
        Default: False

    auth
        Auth string to be used. This can be like 'x-auth-token=XYZ'.
        Type: string

    body
        Body to send with the request
        Type: string

    cookies
        Optional cookies to send with the request.
        Type: object

       ....
```

In Web UI control panel there is no any ordering and it's a bit confusing: 
a13fnwj 1

bigmstone:
Fixes #154

### DIFF
--- a/modules/st2-auto-form/auto-form.component.js
+++ b/modules/st2-auto-form/auto-form.component.js
@@ -67,7 +67,6 @@ export default class AutoForm extends React.Component {
         }
       )
       .value();
-    console.log(fields)
 
     return <div>
       {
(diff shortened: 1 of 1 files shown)


armab:
Confirming that feature works :+1:

PR #365, opened 2017-06-22
Allow Re-Run without parameter modification

### LINKED ISSUE #364: Re-run execution with secret parameters from UI is broken
See https://github.com/StackStorm/st2/issues/3339

bigmstone:
Currently in the web-ui we allow for parameter modification when re-running an action. Since the action's parameters are obfuscated when `secret: true` the obfuscated text is sent in place. We should allow the user to re-run an action without parameter modification which would leave the secret parameters untouched. We shouldn't remove the ability to edit parameters when re-running though so this should be optional when re-running an action.

~- [ ] `Edit Parameters` binary switch~
- [x] Logic to include parameters when ~`edit` binary is `true`~ param is modified.
- [x] Tests

Fixes #364 

### Update:
After toying around with this the way I've implemented seemed to be the right way to go. Basically the workflow is exactly as it used to be, except with we evaluate which params have been modified and only include those in the payload. So if a secret is left alone it is not included and the old param is honored. If a secret field is modified it is included in the payload. This got around an issue I couldn't get over which was "What if someone wants to send ******** as their actual string?". Admittedly a corner case, but this allows them to. This also doesn't change any of the existing logic so it's "mostly" covered by existing tests.

### DIFF
--- a/apps/st2-history/history.controller.js
+++ b/apps/st2-history/history.controller.js
@@ -266,7 +266,8 @@ module.exports =
       },
 
       submit: function (parameters) {
-        st2api.client.executions.repeat($scope.record.id, { parameters }, {
+        console.log(parameters);
+        st2api.client.executions.repeat($scope.record.id, { }, {
           no_merge: true
         })
           .then((record) => {
--- a/apps/st2-history/history.controller.js
+++ b/apps/st2-history/history.controller.js
@@ -266,8 +266,7 @@ module.exports =
       },
 
       submit: function (parameters) {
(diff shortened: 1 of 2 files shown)


Kami:
Yep, agreed.

The way it works in the API is that user only needs to provide parameters which are to be overridden and they are merged with original parameters inside the API.

This means we could still allow user to override parameters, but only only if value for secret parameter is not `****************` or similar.

Kami:
LGTM, but would also need to test it locally :P

PR #373, opened 2017-07-12
Remove auth tokens from query params and allow redefine Stream url 

enykeev:
Requires https://github.com/StackStorm/st2/pull/3566

### DIFF
--- a/modules/st2-api/api.service.js
+++ b/modules/st2-api/api.service.js
@@ -49,10 +49,26 @@ module.exports =
             prefix: auth.path
           };
         }
+
+        if (server.stream && _.isString(server.stream)) {
+          var stream = URI.parse(server.stream);
+
+          if (stream.port && !stream.hostname) {
+            stream.hostname = window.location.hostname;
+          }
+
+          opts.stream = {
+            protocol: stream.protocol,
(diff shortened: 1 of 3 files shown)


enykeev:
Ok, as a result of miscalculation on my part, this PR gets (at the very least) delayed.

The problem is that by default browsers refuse to accept cookies received cross-origin. While in theory we can solve this problem by redirecting user to the proper origin during authentication routine, it would create a fiddly user experience and may introduce other security-related problems potentially more dangerous than the one we're currently trying to fix.

christopherjbaker:
@enykeev Does this apply to the new react UI? And, if so, should we open an issue to fix it, as I assume the changes here certainly don't apply.

enykeev:
Yes, this change needs to be transitioned manually. Problem is, even though I've implemented it, I'm still not sure if this is the right way to handle the problem. Replacing query parameter with cookie opens us up to XSS and makes it so much more important to properly configure CORS. And if there's one thing this project taught me over the years, it that people simply don't get CORS and there's no amount of explaining or hand holding that can make up for it being poorly designed to begin with.

enykeev:
Re-implemented as a part of #530.

PR #377, opened 2017-07-14
Positional Param Sorting

### LINKED ISSUE #367: Action parameter order must be as per the 'position' attribute for both GUI and CLI
Position field is used to specify parameter order in actions until https://github.com/StackStorm/st2web/pull/363 in 2.3.  

If the position field is populated in the metadata,  it must take the precedence for both CLI and UI.

bigmstone:
Fixes #367 

To make more consistent with st2cli and to solve problem for suites we've move sorting into this order:

1) Required or not
2) Position
3) Alphabetical

### DIFF
--- a/modules/st2-auto-form/auto-form.component.js
+++ b/modules/st2-auto-form/auto-form.component.js
@@ -61,6 +61,15 @@ export default class AutoForm extends React.Component {
       .sort(
         (a, b) => {
           if(a.required === b.required){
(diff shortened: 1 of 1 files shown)


enykeev:
Frankly, I'm lost in this logic. Any idea how to make it more readable?

bigmstone:
I was hoping you'd have some suggestions as I really hate how this looks. 🙁

PR #378, opened 2017-07-14
Updating shrinkwrap to 1.0 compatible version

bigmstone:


enykeev:
In general, I'm all for proper shrinkwrap, But I haven't seen this particular format before. What version of NPM does that?

bigmstone:
I believe it's >= 5.0

bigmstone:
Also I believe it's moving to `package-lock.json` now.

PR #380, opened 2017-07-20
Adding Timezone Support

### LINKED ISSUE #292: Support for timestamp display timezones
st2web part of this change - https://github.com/StackStorm/st2/issues/2532

bigmstone:
This PR should serve as a launch pad for a conversation on how we want to handle TZ as well as implementing what gets decided.

Fixes #292


enykeev:
I tried to approach this issue previously. The two main problem are:
 - that we don't have a place in UI for per-user defined settings (and at the same time, don't really have a lot of this settings to justify space for them)
 - that the list of possible timezones is so massive, we would need some form of lazy loading for it. Besides, I wasn't able to find the complete list of TZs a browser supports. My guess that there is no single standard and browsers are still implementing it as they pleased.

bigmstone:
@enykeev agreed on both points. This PR was the laziest way I could accomplish the ask once I got my bearings and get a covo going.

If we want the ability to modify then I have two options in my head:
1) When you click the user there's a dropdown there w/ a list of available timezones.
2) Have a more formal modal for "settings", again w/ the modal toggle available under the user dropdown

I'm not opposed to using something like `react-timezone` unless you have good reason to avoid it for the dropdown.

armab:
My understanding is that there should be a central `st2.conf` setting for the time zone and all other clients (`CLI`, `WebUI`) should honor that.

What matters: when specific action happened on specific server. And that server even can have wrong time (`ntpdate`), but the idea is to follow that single point of truth.
Having every CLI/WEB/ST2 their "special" way of recognizing the TZ/time can bring even more confusion.

bigmstone:
I'm actually curious now why we have a configureable timezone in `st2.conf` and don't honor locally configured TZ on the server? Timezone is usually relevant per user. So if a user is configured to use CST, PST, etc we should display the time in that fashion, IMO. I get having a single source of truth, but wouldn't it make more sense to have the server's source of truth for logging etc and have a user configurable timezone for CLI/Web UI?

armab:
I just read a bunch of old Issues related to Time Zones we had. Looks like indeed we need to make happy everyone :) Based on that:

Server -> st2.conf (if exists).
And after that honor custom configurations for CLI or st2web if they set (local storage/long cookie or etc).

That'll give an option for everyone.

bigmstone:
To add more fuel to the fire it would appear that the only time `local_timezone` is used is in `st2reactor`. Not for logging, history, etc. The `~/.st2/config` option is only for `st2client`. The central setting for TZ appears to not, by design, have anything to do with UI/UX timezone.

I even went as far as to implement a majority of the API endpoint to get `st2config` before I got enraged by `oslo`. That digging is what lead me to see that the setting for timezone is only used by reactor. I think I'm more confused now then when I started.

We could go one of two ways from here - IMO (Unless my deduction from above is totally off base, which is possible). We could go route 1. Where we make a big push to unify how we handle TZ - though what we'd actually do here is unclear to me. Route 2. is to just implement st2client style TZ handling in st2web where the user is given a TZ option and that is set in a cookie.

bigmstone:
Per slack convo w/ @lakshmi-kannan and @Kami we're going route 2. Implement `st2client` style TZ handling and allowing user to select timezone. I will have it default to local timezone if no cookie is set and local timezone is available.

Kami:
> Per slack convo w/ @lakshmi-kannan and @Kami we're going route 2. Implement st2client style TZ handling and allowing user to select timezone. I will have it default to local timezone if no cookie is set and local timezone is available.

Sounds good. Only thing I might consider doing differently is perhaps defaulting to UTC instead of a local user timezone.

apps/st2-actions/template.html
```diff
@@ -1,7 +1,7 @@
 <header class="st2-menu"></header>
 
 <main class="st2-panel">
-
+  <div class="st2-redux"></div>
```
enykeev: Do we still need redux?

apps/st2-history/template.html
```diff
@@ -2,6 +2,7 @@
 
 <main class="st2-panel" data-test="history_panel">
 
+  <div class="st2-redux"></div>
```
enykeev: Same.

apps/st2-history/template.html
```diff
@@ -359,7 +372,11 @@ <h2 class="st2-details__panel-title">Trigger Details</h2>
               </dt>
               <dd class="st2-details__panel-body-value"
                   ng-if="record.trigger_instance.occurrence_time">
-                {{ record.trigger_instance.occurrence_time | date:'EEE, dd MMM y HH:mm:ss UTC':'UTC' }}
+                <div class="st2-time"
+                  timestamp="record.trigger_instance.occurrence_time"
+                    utcdisplay="utcDisplay"
```
enykeev: Mixed indent

modules/st2-menu/menu.directive.js
```diff
@@ -11,7 +11,7 @@ module.exports =
       templateUrl: template,
       link: function postLink(scope) {
         scope.routes = scope.state.get().map(route => {
-          return Object.assign(route, {
+         return Object.assign(route, {
```
enykeev: Messed up indent

modules/st2-time/time.component.js
```diff
@@ -0,0 +1,32 @@
...
+  let {format, timestamp, utcdisplay, togglecallback, datatest} = props;
+  let dateFormat = format ? format : 'ddd, DD MMM YYYY HH:mm:ss';
+  let timeStamp = new Date(timestamp);
+  if (utcdisplay){
+    dateFormat += ' UTC';
+  }
+
+  return (
+    <Time
+      data-test={datatest ? datatest : undefined}
```
enykeev: I'd rather you would use {...restProps}. `Data-` props are falling through as is in react.

package.json
```diff
@@ -13,19 +13,24 @@
     "angular-ui-notification": "0.1.0",
     "angular-ui-router": "1.0.0-alpha.1",
     "lodash": "2.4.2",
-    "ngreact": "0.2.0",
+    "ngreact": "^0.4.1",
```
enykeev: Please stick to exact versions. I recommend to set it once with `npm config set save-exact true`.

LindsayHill:
Adding some comments made on Slack:

I did some brief testing of this in my lab
```
lhill@st2:~$ npm -v
3.10.10
lhill@st2:~$ node -v
v6.11.2
lhill@st2:~$ st2 --version
st2 2.3.2
lhill@st2:~$ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 14.04.5 LTS
Release:        14.04
Codename:       trusty
lhill@st2:~$
```
Generally looks OK, but I noticed that the TZ for child tasks was not updating. It always displayed local time, not UTC. Clicking on timestamp in parent task would change timezone, but no change for child tasks. See 
st2web_tz_crop

bigmstone:
@LindsayHill sadly I can't repro this. Can you provide system info/etc. And any console errors you're seeing?

bigmstone:
@enykeev Unfortunately using spread to get remaining props would't work in this instance since `ngreact` isn't sending anything w/ `-` in the key. I couldn't find any documentation on it - specifically if it's an `angular` problem or an `ngreact` problem. When I started parsing the `ngreact` code base I decided it wasn't worth the effort for what I'm doing here.

LindsayHill:
👍 
Now looks good to me, with sub tasks operating as expected

package.json
```diff
@@ -59,9 +62,11 @@
     "gulp-uglify": "1.5.3",
     "gulp-util": "3.0.7",
     "gulp-webserver": "0.9.1",
+    "moment": "2.18.1",
     "ngify": "1.4.1",
     "npm-shrinkwrap": "200.5.1",
     "react-addons-test-utils": "0.14.8",
+    "react-time": "4.3.0",
```
enykeev: Yep, missed that one. Both `react-time` and `moment` are supposed to be dependencies, not devDependencies. Will change it in my PR.

PR #385, opened 2017-07-26
Updating Packages

bigmstone:


enykeev:
Any particular reason to do that now? Pack UI will going to introduce new dependencies and in some time will help us get rid of others. I would prefer not to shake the boat in the process.

Also, the packages I've pinned with git urls may include the changes that still haven't been merged to upstream. Updating them blindly may have consequences.

PR #386, opened 2017-07-26
Revert "Updating shrinkwrap to 1.0 compatible version"

enykeev:
The commit resulted in a bunch of local issues related to incompatibilities between outdated packages and a fresh version of node and npm.

My suggestion is to revert back to shrinkwrap v0 until the next refactoring phase.

Also, since @bigmstone joined me on this frontend endeavor, it makes sense for us to agree to use Node v6 and NPM v3 for now. Also, to make diffs more readable, I suggest using `gulp shrinkwrap` instead of `npm shrinkwrap` and manually removing `from` field from the file with the handy regex `^\s+"from": "\S+",\n`. I know this seems like a chore, I'm really hoping to get through that soon.


bigmstone:
@enykeev will you also update `package.json` `engines` to reflect this version?

PR #388, opened 2017-07-31
Use pack.ref to get pack icon via API

lakshmi-kannan:
Packs API now uses pack ref as opposed to name
https://github.com/StackStorm/st2/blob/master/st2api/st2api/controllers/v1/packviews.py#L175

But UI was still using pack.name to get the icon. For some of the packs like Circle CI, name is a more user friendly "Circle CI" and ref is "circle_ci". So we should really be using ref to get the icon. 

Pack model defines "ref" as mandatory. So we can assume packs that are being registered will always have "ref" https://github.com/StackStorm/st2/blob/master/st2common/st2common/models/db/pack.py#L43. 

## After

Actions

Rules

### DIFF
--- a/apps/st2-actions/actions.controller.js
+++ b/apps/st2-actions/actions.controller.js
@@ -78,8 +78,8 @@ module.exports =
           $scope.icons = {};
           _(packs).forEach(function(pack) {
             if (pack.files && pack.files.indexOf('icon.png') >= 0) {
-              var icon_path = st2api.client.packFile.route(pack.name+'/icon.png');
-              $scope.icons[pack.name] = icon_path;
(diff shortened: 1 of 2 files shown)


lakshmi-kannan:
BTW we infer the pack name from actions and rules it looks like. Therefore we don't display the user friendly name in the pack titles. We only display the ref which is sad. To fix this, the UI should cache a ref to name lookup for packs and then use the display name for packs as opposed to ref. That is a separate fix.

Kami:
Good catch :+1: 

I believe I was already fixing something like that in the past, but it could be a more recent regression :)

PR #396, opened 2017-08-15
Fixing position sort

bigmstone:
Sorting by position is currently broken because the effective check of:
```
if (0){
}
```

Does not not pass. So we must check if position is explicitly undefined so that we will branch into the sort logic even if the position value is 0.


LindsayHill:
Tested, LGTM

Issue #417, opened 2017-08-30
Edit rule breaks over adding criteria


dzimine:
When I add a second criteria to the rule, it REPLACES the first criteria instead of adding to it.

christopherjbaker:
This has been fixed in the [react rewrite](https://github.com/StackStorm/st2web/pull/420), which should be included with 2.6.

Issue #415, opened 2017-08-26
Pack app "Preview" does nothing


dzimine:
Not even clear how it is intended to work, but as of now, just doesn't do anything. 
`st2 v 2.4.0`

![image](https://user-images.githubusercontent.com/1294734/29742480-e426e6a2-8a34-11e7-8ac8-95abceb58362.png)

bigmstone:
@dzimine I can't reproduce this issue (Chrome + macOS) - though I'll admit if you don't know to scroll down it appears as though nothing happened. Can you confirm this isn't working or that you did not scroll down?

If it's in fact not working which browser/os combo are you using?

rakhesha:
Hi, I too facing the similar issue (StockStorm Version: st2 2.5.0 + Windows 8.1 + Chrome), Is there any particular cause for this issue? Unable explore the "Preview" button. Suggestions are welcome.

christopherjbaker:
This has been fixed in the [react rewrite](https://github.com/StackStorm/st2web/pull/420), which should be included with 2.6.

Issue #414, opened 2017-08-23
Workflow and status icon alignment on the History tab needs to be fixed


samirsss:
The alignment of the status icon for workflows is out of sync with the status for individual actions in the history tab. This seems to be something recent. I am on the latest 2.4 builds for ST2

![screen shot 2017-08-23 at 9 31 21 am](https://user-images.githubusercontent.com/7357878/29627152-f36dd3d8-87e5-11e7-856e-352590e1043a.png)

vincent-legoll:
I'm on 2.4.1 / centos7 and don't see that.

My workflows' status icons are at the same place as they are for single actions 

Maybe it has been fixed.

armab:
It also makes sense mentioning the browser and its version to have instructions to reproduce the bug.

christopherjbaker:
This has been fixed in the [react rewrite](https://github.com/StackStorm/st2web/pull/420), which should be included with 2.6.

Issue #381, opened 2017-07-20
CI Should fail on Lint Errors


bigmstone:
If CI throws a lint error it should fail. It currently passes. This is primarily for my PRs. 😄

enykeev:
Yeah, I've noticed some linting errors, but was myself under the impression they are failing. I'll take a look today, there's something in `gulp` that prevents that.

bigmstone:
@enykeev I clean up those linting issues in #377

christopherjbaker:
This has been fixed in the [react rewrite](https://github.com/StackStorm/st2web/pull/420), which should be included with 2.6.

PR #420, opened 2017-09-08
Angular to React transition

enykeev:
Modules left to transition:
- [x] ActionReporter
- [x] Criteria
- [x] Filter
- [x] Flow Link (look into making Flow an st2web app instead)
- [x] History Flex Cald
- [x] HistoryChild Flex Cards
- [x] Rule Flex Card
- [x] Notifications (implemented, just need to be its own module)
- [x] Proportional (as an option, make parameters row in history flex card to be horizontally scrollable)
- [x] Remote Form
- [x] Scroll into view
- [x] View

In addition to that (during that PR or later):
- [x] split modules into separate npm package (see [lerna](https://lernajs.io/))
- [ ] reimplement build system to make it easier to use (albeit more opinionated, see [st2apidocgen](https://github.com/StackStorm/st2apidocgen) or [react-scripts](https://github.com/facebookincubator/create-react-app/tree/master/packages/react-scripts))
- [ ] move from less to postcss, switch to css modules
- [ ] see how we can leverage [Chromeless](https://github.com/graphcool/chromeless)
- [ ] look into separating code view into its own app

Additional Tasks (primarily cleanup and consistency related; not required for release):
- [x] organize imports
- [x] modules use `.isRequired`
- [x] don't use props = {}, put on tag directly
- [x] all modules support className and ...props (make sure to exclude all props)
- [x] unpin all dependencies to carat
- [x] apps handle 0 returned results
- [x] use classnames module
- [x] separate `st2-apps/<app>/<app>-details.component.js`
- [x] apps use `urlParams` as source of info
- [x] integrate notifications
- [x] page titles
- [x] remove instances of ngModel


apps/st2-history/history-panel.component.js
```diff
@@ -79,7 +79,7 @@ export default class HistoryPanel extends React.Component {
           store.dispatch({ type: 'SELECT_EXECUTION' });
         }
       })
-      ;
+    ;
```
enykeev: I usually indent `;` too, but I'm open to doing it this way too so long as there is a eslint rule to remind me.

cognifloyd:
@christopherjbaker I'm curious about how extensible this is.
The last several commits deal with data types: Array, Boolean, Enum, Integer, Number, Object, Password, String. These are the types defined by [JSONSchema](http://json-schema.org/documentation.html) as described in the [Action MetaData](https://docs.stackstorm.com/actions.html#action-metadata) docs. How easy would it be to define a custom type? For example, I want to add a module/extension into my instance of st2web that detects all DateTime fields (if it's a rule of type [core.st2.DateTimer](https://docs.stackstorm.com/rules.html#run-action-on-a-specific-date), then the `trigger.date` parameter is a DateTime field) and adds a calendar + clock widget in place of the text field. I want to specify some other helpful widget/field for certain pack fields as well. So, how would this work?

cognifloyd:
Reading through the code for rules, it looks like the Trigger on a Rule uses RemoteForm to generate an AutoForm based on the triggerSpec which is the schema returned by the api. There's no obvious extension point in there to allow me to mutate the types of particular fields and inject my own widgets. Maybe change the child spec in remote form when the trigger name matches some criteria, and then add an additonal type to AutoForm (somehow) to handle the additional data type?

cognifloyd:
Another example would be a field on a pack that is an "object", but that I want to allow creating that object with a form instead of raw json. I need to somehow inject a custom "type" and "field".

I could see, maybe, adding DateTime as a native type in st2 itself (thus extending JSONSchema), but for such a custom object, I really want to improve the experience of using it in st2web, and that means teaching st2web how to handle my custom object through some kind of extension.

enykeev:
@cognifloyd theoretically, you should be able to "fork" AutoForm, inherit from existing one, add your changes and then inject it back with something like https://www.npmjs.com/package/require-injector. I haven't tried that and extensibility wasn't my main priority (reusability was), but it seem plausible to me and worth a try.

christopherjbaker:
As @enykeev said, the current priority is reusability. That being said, there is no reason that extensibility can't become important. I do think it needs definite discussion and planning, as I have seen a number of projects become unmaintainable when trying to be too extensible. Would you mind creating a feature request, including some of the ways you would like it to be extensible?

enykeev:
Issues I've found:
- when config.js is empty (host section is commented out), the login app returns an error on login attempt. Instead, it should access api and auth on the same host with web ui. This is actually a most commonly used configuration attributing to about 99% of st2 deployments now.
- I don't see any live updates on new executions in either st2history or st2-actions. Seems to me we forget to transition https://github.com/StackStorm/st2web/blob/master/apps/st2-history/history.controller.js#L193-L245 and https://github.com/StackStorm/st2web/blob/master/apps/st2-actions/actions.controller.js#L191-L234 parts.
- If you try to collapse a section in history's list view, it collapses all of them.
- Code views doesn't have a limit on the lines it shows (at the same time, there's no way to open code view in fullscreen)
- History view doesn't parse filters provided as a query param `http://localhost:3000/#/history?action=core.local`. It is needed for st2-actions "See full action history" button to work and it also required for backward compatibility.
- Rule delete button doesn't show the result until the page is reloaded. The rule should be removed immediately after we get a successful response from API.
- Opening http://localhost:3000/#/actions should select first action on the list and expand its pack.

There's likely many more but it should keep you busy for now :)

enykeev:
Also, linting takes way too much time. There's probably a node_modules folder that we didn't properly ignore. Mind taking a look?

christopherjbaker:
Modules:
- [ ] st2-api
- [ ] st2-filter-expandable
- [ ] st2-proportional
- [ ] st2-scroll-into-view
- [ ] st2-select-on-click
- [ ] st2-store
- [ ] st2-style
- [ ] st2-title
- [ ] st2-value-format

Components:
- [ ] st2-action-reporter
- [ ] st2-auto-form
- [ ] st2-criteria
- [ ] st2-filter
- [ ] st2-flex-table
- [ ] st2-forms
- [ ] st2-highlight
- [ ] st2-label
- [ ] st2-login
- [ ] st2-menu
- [ ] st2-notification
- [ ] st2-pack-icon
- [ ] st2-panel
- [ ] st2-popup
- [ ] st2-portion-bar
- [ ] st2-remote-form
- [ ] st2-time
- [ ] st2-view

christopherjbaker:
fixes #432, #429, #424, #423, #417, #415, #414, #381, #375, #253

christopherjbaker:
When circleci runs, it uses lerna to compile against the local modules folder, and creates the packages based off this. I think the best option when building packages would be to use the npm packages instead of the local copies (and run the tests against them, as well). This would ensure that the published modules are also up to date (and, I think, at least marginally reduce the build size).

@enykeev Would it be unreasonable for `v0.0.0` branches to use npm instead of lerna?

enykeev:
> This would ensure that the published modules are also up to date (and, I think, at least marginally reduce the build size).

I don't see how it can reduce build size, but it can improve build time.

Nevertheless, I don't think that's a good idea simply because the fact that tests are passing doesn't necessarily saying that the module in npm matches the code in the repo. The problem you seem to try to solve here (all the recent changes are published):

- may not really be a problem (wip changes you're not ready to publish)
- better solved by comparing integrity hash of the package and its dependencies
- in some ways solved by npm by comparing integrity during package install
- could (and probably should) be checked on timer rather than on build

> Would it be unreasonable for only v0.0.0 branches to deploy, and for those to use npm instead of lerna?

If you're talking about deb\rpm, `master` and `v` branches are deployed in the separate repos. Master is deployed to unstable staging while version branches are deployed to stable staging. If, on the other hand you're talking about module deployment (publishing), I think that module versioning should be independent of st2web versioning mainly because the latter suffers from marketing rather than technical approach to versioning which is unhealthy to module ecosystem.

enykeev:
On the user-facing side of things, it is mostly looking good. There are few things I found that bothers me:

1) Opening new app sometimes results in "no results were found" message until the request is complete. I'd rather we don't show anything until the state is complete.

2) In history, Rule Details link points to `/rules/<name>/general`, but we don't have a handler for `/general` part, so it defaults to `/rules`. We can change the link, but I'd rather we add the handler in case user has this link in bookmarks.

3) Action input and output sections in history details should tell us if the section is empty due to history record not having said properties to begin with (as it turns out with core.noop) rather than as a result of an error.

4) Rerunning a core.local action results in error:
```
Uncaught (in promise) TypeError: Cannot read property 'stdout' of undefined
    at debug (run-local.js:7)
    at ActionReporter.render (action-reporter.component.js:26)
    at finishClassComponent (react-dom.development.js:7873)
    at updateClassComponent (react-dom.development.js:7850)
    at beginWork (react-dom.development.js:8225)
    at performUnitOfWork (react-dom.development.js:10224)
    at workLoop (react-dom.development.js:10288)
    at HTMLUnknownElement.callCallback (react-dom.development.js:542)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:581)
    at invokeGuardedCallback (react-dom.development.js:438)
```

christopherjbaker:
@enykeev My history rule link (with `/general`, as we do have a handler for this) is working properly. Can you look into more of why it's not working on your end?

enykeev:
@christopherjbaker yeah, sorry, I checked it again and figured that the rule I was trying to access have been deleted and that's why it routed me back to `/rules`. I guess we need some kind of indication for such cases, though, certainly not as part of this PR.

enykeev:
I can confirm that the problem I've described yesterday have all been fixed.

I however can also confirm that the problem @bigmstone have mentioned in https://github.com/StackStorm/discussions/issues/280 that related to create rule button is still present and is caused by only the plus sign being a click handler rather than the whole button. There's also some button alignment problem in the create rule popup itself.

I also have a problem with a cancel button in history details. It acts as if it's not wired anywhere at all.

armab:
Wohoo, Congrats! :tada:

PR #426, opened 2017-09-19
Enum default value

### LINKED ISSUE #425: Default value not honored on enum
Brought from StackStorm/st2#3205 via @pietervogelaar: 

In addition to #2889, the default parameter value isn't selected in the selectbox when enum is used in the UI.

bigmstone:
Fixes StackStorm/st2#3205 and fixes #425

### DIFF
--- a/modules/st2-auto-form/auto-form.component.js
+++ b/modules/st2-auto-form/auto-form.component.js
@@ -59,6 +59,19 @@ export default class AutoForm extends React.Component {
     return onChange && onChange(name, value);
   }
 
+  componentWillMount(){
+    const { spec } = this.props;
+
+    if (_.has(spec, 'properties')){
+      Object.keys(spec.properties).forEach(function(key) {
+        let value = spec.properties[key];
+        if (value.default !== undefined && value.required === true){
+          this.handleChange(key, value.default);
+        }
+      }, this);
+    }
+  }
+
   render() {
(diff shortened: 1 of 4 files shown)


bigmstone:
Currently this only takes into account required params with a default value. Should we include all values that contain a default? /cc @lakshmi-kannan @enykeev

LindsayHill:
> Currently this only takes into account required params with a default value. Should we include all values that contain a default?

Guess that depends what happens if I don't select anything for those params. My expectation would be that it does not send anything, and that ST2 will automatically use the default - same as if I ran it via CLI, and did not specify anything for that parameter. I think we made some changes to this on the backend recently. I think previously it would *not* use a default value for optional params, but this has changed.

bigmstone:
So to recap from @ramgolla @samirsss @lakshmi-kannan and I's convo on slack:

1) There's not currently a bug w/ how it works when there's a default value set.
2) The current UX is a bit confusing because if the default param in enum isn't the first value it isn't "selected"
3) The current UX is a bit confusing because if there is no default param the first selection in the enum is still selected - despite the fact that it will not populate that value in the API call.

To improve the UX we will have the dropdown's first item be empty if there is no default value. If there is a default value it will auto-select the correct value in the enum (currently implemented in this PR).

ramgolla:
+1

m4dcoder:
LGTM.

PR #427, opened 2017-09-27
Generate URL to api/auth based on current URL

shusugmt:
When `host` parameter is not set in `config.js` which is by default, st2web will generate url to api/auth endpoint with fixed static `https` protocol and port 443. This behavior makes it a bit complicated when using `st2-docker` image:

- We need to map container port 443 to host 443 port
- or, if we really want to change, then we need to provide custom `config.js` that points to the correct api/auth URL with a mapped port number (like below)

```
'use strict';
angular.module('main')
  .constant('st2Config', {

    hosts: [{
      name: 'default',
      url: 'https://:30443/api',
      auth: 'https://:30443/auth'
    }]
  });
```

This PR will fix this by simply using `window.location` object for protocol and port number to generate endpoint URLs.

Note: I avoided using `window.location.origin` property because of lack of browser support (IE10)

### DIFF
--- a/modules/st2-api/api.service.js
+++ b/modules/st2-api/api.service.js
@@ -51,8 +51,8 @@ module.exports =
         }
       } else {
         opts = {
-          api: 'https://' + window.location.hostname + ':443/api',
-          auth: 'https://' + window.location.hostname + ':443/auth',
+          api: window.location.protocol + '//' + window.location.host + '/api',
+          auth: window.location.protocol + '//' + window.location.host + '/auth',
           token: !_.isEmpty(token) ? token : undefined
         };
       }
--- a/modules/st2-api/api.service.js
+++ b/modules/st2-api/api.service.js
@@ -51,8 +51,8 @@ module.exports =
         }
       } else {
         opts = {
-          api: window.location.protocol + '//' + window.location.host + '/api',
-          auth: window.location.protocol + '//' + window.location.host + '/auth',
+          api: 'https://' + window.location.host + '/api',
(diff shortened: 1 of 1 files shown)


enykeev:
Thanks for the contribution! The only problem I see there is the fact that user may end up with trying to connect to st2auth by http exposing his password in plain text. My opinion is that we should hardcode https protocol for user's sake.

Also, we're doing massive overhaul of the internal architecture of st2web at the moment. Please remind me so I don't forget to port this feature to #420 or create another PR to this branch.

shusugmt:
> The only problem I see there is the fact that user may end up with trying to connect to st2auth by http exposing his password in plain text. My opinion is that we should hardcode https protocol for user's sake.

That should be handled by httpd that serves st2web assets, IMHO. But I also understand your point and what I care the most here is the port number. So I'm pretty fine with simply switching  `window.location.protocol` part back to static `https` string. Let me know if it's OK with you, I'll push a fix.

> Please remind me so I don't forget to port this feature to #420 or create another PR to this branch

OK I'll take a look at it.

FYI I found this PR is also related:
https://github.com/StackStorm/st2web/pull/373/files#diff-6e01c44c310854205e63e36ef07c7729R71

enykeev:
> That should be handled by httpd that serves st2web assets

In production, sure, but there is still a huge room for interpretation and lack of hindsight. Password that is exposed once, can never be trusted.

>  Let me know if it's OK with you, I'll push a fix.

Go for it.

shusugmt:
Fix pushed.

PR #460, opened 2018-01-12
st2-criteria: accept patterSpec (fixes #219)

### LINKED ISSUE #219: Hints for nested properties in Sensor `payload_schema` (house.dog.fleas)
The cool thing I found (suddenly :facepunch:) is that Sensor payload schema supports `description` for each property and even more, webui shows the hints!

That's just awesome, however never seen someone uses that in any docs or examples.

----

The cool thing will be also to show hints for every child property in sensor `payload_schema`, like `house.dog.fleas`

christopherjbaker:


christopherjbaker:
@enykeev

PR #461, opened 2018-01-12
Fix issues uncovered by st2cicd

enykeev:


modules/st2-auto-form/auto-form.component.js
```diff
@@ -54,7 +54,7 @@ export default class AutoForm extends React.Component {
 
   handleChange(name, value) {
     const { spec, data, onChange } = this.props;
-    return onChange({
+    return onChange && onChange({
```
christopherjbaker: While this removes the error, it masks the underlying issue. As this is a controlled component, it must have an onChange value.

Unfortunately, it can also be used without onChange if disabled is set to true, which is why it doesn't have isRequired. I have created a [PR](https://github.com/StackStorm/st2web/pull/463) that adjusts the propTypes accordingly.

christopherjbaker:
The `onChange` error that you posted to slack follows. Is the CI using the latest dev package? The error references `Autoform.componentWillMount`, which was [adjusted](https://github.com/StackStorm/st2web/commit/df016741a7fc15b7aae1d71de943190b695e83cd#diff-128fae3a76a671248d8594578026ceea) a few commits ago.

```
auto-form.component.js:72 Uncaught (in promise) TypeError: onChange is not a function
    at AutoForm.handleChange (auto-form.component.js:72)
    at AutoForm.<anonymous> (auto-form.component.js:37)
    at Array.forEach (<anonymous>)
    at AutoForm.componentWillMount (auto-form.component.js:34)
    at callComponentWillMount (react-dom.development.js:6370)
    at mountClassInstance (react-dom.development.js:6435)
    at updateClassComponent (react-dom.development.js:7840)
    at beginWork (react-dom.development.js:8225)
    at performUnitOfWork (react-dom.development.js:10224)
    at workLoop (react-dom.development.js:10288)
```

christopherjbaker:
I compared all the api requests from the old app and the new and found a few more parameter differences.

PR #464, opened 2018-01-13
st2-label: colorblind support (fixes #439)

### LINKED ISSUE #439: colorblind support
On some screens or setups it might be fairly difficult for colorblind people (myself included in this example) to see the difference between the green/red success status dot. Changing the colors to a more contrastful tones of green/red would be helpful but maybe adding an icon in the middle of the dot would be ideal:
image

This second option might be detrimental to screen readers, but maybe a middle ground could be found.

christopherjbaker:


christopherjbaker:
Closing as this will be handled with @adrifolio's UX updates.

PR #493, opened 2018-01-28
Fix hardcoded port number to auth/api endpoint URL

shusugmt:
There's a regression in 2.6 and some users of docker 1ppc image are facing the problem that they can't login to web UI if that instance is not using port 443 to expose st2web service (which is default for sample kubernetes setup).

This patch will fix it by doing the exact same as https://github.com/StackStorm/st2web/pull/427

### DIFF
--- a/modules/st2-api/api.js
+++ b/modules/st2-api/api.js
@@ -65,8 +65,8 @@ export class API {
     }
     else {
       opts = {
-        api: `https://${window.location.hostname}:443/api`,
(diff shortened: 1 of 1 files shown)


enykeev:
Well, that's embarrassing. @shusugmt thanks for the contribution. Again. We should probably add a test for that. Mind taking a stab at it?

christopherjbaker:
@shusugmt I merge this into a branch with tests, #497.

PR #495, opened 2018-02-06
Add support for advanced filters

enykeev:
screen shot 2018-02-06 at 12 44 04

### DIFF
--- a/apps/st2-history/history-panel.component.js
+++ b/apps/st2-history/history-panel.component.js
@@ -22,6 +22,7 @@ import {
   PanelNavigation,
   Toolbar,
   ToolbarFilters,
+  ToolbarSearch,
   ToolbarView,
   Content,
   ContentEmpty,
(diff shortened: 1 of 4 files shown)


apps/st2-history/history-panel.component.js
```diff
@@ -157,9 +160,26 @@ export default class HistoryPanel extends React.Component {
...
+    if (!tokens.length) {
+      return;
+    }
+
+    for (const token of tokens) {
+      const [ , v ] = token.split(':');
+      if (!v) {
+        return;
+      }
     }
```
christopherjbaker: should we add a comment for what this is looking for? It looks like it will halt requests until the filter has a value, so things like `foo` and `foo:` will be blocked. Might be pertinent to move this logic to a `isFilterValid` method, making it read a little cleaner, especially with future code that might have a more rigorous check here.

Should we add some throttling (somewhere), so that `foo:bar` will not trigger 3 lookups?

apps/st2-history/history-panel.component.js
```diff
@@ -168,14 +188,15 @@ export default class HistoryPanel extends React.Component {
     this._source.removeEventListener('st2.execution__delete', this._executionDeleteListener);
   }
```
christopherjbaker: perhaps a `fetchGroupsThrottled` method?

Issue #500, opened 2018-02-27
UI refreshes and previous position of the browser window is lost, returning to the top of the page


jbrownsc:
When watching actions run in st2web UI, the position of the browser window is constantly being returned to the top of page.  If I have a long list of actions running, this requires me to wait until all actions have completed to scroll down.

This is on Chrome 64.0.3282.167 and 58.0.2 (64-bit)

enykeev:
I've temporarily disabled the behavior in #502. It didn't quite make it through the transition from Angular to React and will require a bit more effort. I'll keep the issue open until it could be properly fixed, but the behavior should not cause that much trouble anymore.

LindsayHill:
st2web.2.6.0-5 has been published to stable repos - try that

jbrownsc:
I've been using st2 2.7.1 and this issue appears to be resolved!  Nice work!

PR #502, opened 2018-03-01
Temporary disable scroll into view for executions

enykeev:
The proper fix itself seems quite involved so I've decided to disable it for now until I can came up with the simpler solution.


christopherjbaker:
Yes. #500

PR #518, opened 2018-04-24
fix quickstart (use lerna command.)

kaishuu0123:
Before the change was written as npm install.
Is it correct to use the lerna?
Or is there any other document to be referred?


enykeev:
Thanks for the contribution!

PR #521, opened 2018-05-04
Fix Triggers render

enykeev:


armab:
Is there any issue/bug this PR solves?
Would be nice to cross-link then or include a description what this PR really fixes.

enykeev:
No, this is something I broke during #499 and fixed before anyone would notice.

armab:
Alright, thanks!

PR #522, opened 2018-05-04
Update Packs layout

enykeev:
screen shot 2018-05-04 at 8 17 23 pm


cognifloyd:
Having a bigger details panel for the packs looks so much more usable!

Issue #510, opened 2018-03-22
Proposal: Remove Reamaze integration, add Forum.stackstorm.com


LindsayHill:
Our new [forum](https://forum.stackstorm.com) is coming along nicely.

I think we should remove the Reamaze integration from st2web, and instead have a link to forum.stackstorm.com. 

I would like to do this in v2.7.

Thoughts?

christopherjbaker:
Might be too much, but would probably be possible to integrate a small
search in a panel.

LindsayHill:
That might work. MVP is just to link to forums, but if there's better integration, then we should look at it.

Main thing is that if we do something it doesn't block on systems that don't have direct Internet access.

Kami:
:+1:  for integrating that into WebUI.

As long as it's unobtrusive integration (link is fine) - aka only loads on request or similar so discourse doesn't track every user...

LindsayHill:
Resolved by #527

PR #527, opened 2018-05-25
Replace Reamaze integration with link to the forum

enykeev:


LindsayHill:
👍 

Resolves #510

modules/st2-menu/menu.component.js
```diff
@@ -118,8 +118,8 @@ export default class Menu extends React.Component {
           Docs
         </a>
 
-        <a href="mailto:support@stackstorm.com" className="st2-menu__side" data-reamaze-lightbox="contact">
```
armab: How this change affects `bwc-ui`?

I agree that it's good idea to remove support email for Enterprise version, but I think it makes sense to leave it in Enterprise UI.

modules/st2-login/login.component.js
```diff
@@ -207,8 +207,8 @@ export default class Login extends React.Component {
             <a target="_blank" rel="noopener noreferrer" href="http://docs.stackstorm.com">
               Documentation
             </a>
-            <a href="mailto:support@stackstorm.com" data-reamaze-lightbox="contact">
```
armab: It's funny how the PR title advertizing "Replace Reamaze integration", but in fact removes also Support Email.

modules/st2-menu/menu.component.js
armab: How this change affects `bwc-ui`? Was the decision to remove Support Email for enterprise version as well?

I agree that it's good idea to remove support email for Community version to reduce number of free-loaders, but I think it makes sense to leave it in Enterprise UI to have a differentiator in support between versions.

enykeev: I didn't "remove" email the way you think I did. Mailto was always a fallback in case reamaze is not working for some reason. Now that we have a link to forum, there's simply no place for this mailto link any longer.

armab: contact-us-forum
Even with mailto fallback, it's still an Email channel.

I see how Forum complements existing support channels, but don't see how Forum replaces "Support via Email" since these channels could be used for different purposes:
- It's OK to ask a st2 technical question at Forum and let Community handle it.
- But one can't post at Forum requests to StackStorm team about Enterprise license terms/pricing/renewal questions or other Enterprise support requests and issues or bugs where we have to react in a timely manner. 

I believe removing Email support channel for Enterprise is the opposite to what we offer as a product: https://stackstorm.com/#product

I think having a differentiator between Community/Enterprise is important for us if we're looking towards sales and do care about the product in general.

This way my opinion: it's totally OK if we remove Email Link for Community version, but would be a regression removing that channel for Enterprise version.

Curious what others think cc @LindsayHill @lakshmi-kannan @dzimine

LindsayHill: My vote:

* st2web: Definitely remove Remaze integration **and** `mailto:` Only have link to forum
* bwc-ui: Definitely remove Reamaze integration, and provide link to forum. Provide `mailto: support@stackstorm.com` link. In future may replace that with link to Extreme support site.

enykeev: Guys, just to be clear, since it's a fallback that only opens the mailto link in case reamaze script have not been loaded for some reason, it is safe to assume that about zero support emails were initiated by pressing "contact us" in the UI. And since we still receive some support requests via email, looks like we have some other ways for users to get to know that we have a support email.

If we want to add email as an additional button to menu, ask Adri to add it to mocks. I'm gonna push back because having two versions of the menu for st2web and bwc-ui will complicate the structure of the app, but hey, if you think it is crucial for us to have this link there, I'll let you do the foot work.

LindsayHill: OK. Remove Reamaze, and don't have any fallback to email. Just have a link to the forum. Customers with support contracts know how to get in touch with support, and other users of that link have been looking for free assistance.

PR #531, opened 2018-06-06
Open docs/forum links in new tab

LindsayHill:
Three related changes here:

1/ Change docs link on login page to use `https:`, save one redirect
2/ Change forum link on login to open in new page, same as docs link
3/ Change forum & docs link on main page to open in new page

### DIFF
--- a/modules/st2-login/login.component.js
+++ b/modules/st2-login/login.component.js
@@ -203,10 +203,10 @@ export default class Login extends React.Component {
           </LoginRow>
 
           <LoginBottomRow>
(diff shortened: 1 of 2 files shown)


LindsayHill:
I realized there's some overlap with what @enykeev is doing here https://github.com/StackStorm/st2web/pull/530/files#diff-33c3b98819f1fc216a2d0258f758ea35R39

Will wait for his input, probably best to just combine these as one, rather than deal with merge conflicts.

enykeev:
LGTM. Just merge it to `bump-prerelease` branch instead

PR #562, opened 2018-06-19
Fix login supporting multiple servers

enykeev:


LindsayHill:
Thanks, tested it out and it works for me

PR #593, opened 2018-07-20
Set request content type

enykeev:


armab:
As noted in Slack (https://stackstorm.slack.com/archives/C029K4CBU/p1532086199000268) I've tried the packages from this PR and tested with developer console, - the headers were not set yet.

So could be something else going on.

PR #596, opened 2018-07-25
Code app

enykeev:
A separate application for viewing different kinds of code fragments: entities (Actions, Rules, Executions), entry_point files, execution results and trigger instance payloads.

Previously, we showed them as a full screen pop-up, but since we now want to be able to follow live updates of execution results, the task becomes too heavy for the st2-highlight module itself.

At some point, we may want to even add editor to this app and allow user to edit code directly inside the browser, but this certainly lies outside of scope of this PR.

### DIFF
--- a/apps/st2-actions/actions-details.component.js
+++ b/apps/st2-actions/actions-details.component.js
@@ -254,7 +254,7 @@ export default class ActionsDetails extends React.Component {
         { section === 'code' ? (
           <DetailsBody>
             <DetailsPanel data-test="action_code">
-              <Highlight code={action} />
+              <Highlight code={action} type="action" id={action.ref} />
             </DetailsPanel>
           </DetailsBody>
         ) : null }
--- a/apps/st2-actions/actions-details.component.js
+++ b/apps/st2-actions/actions-details.component.js
@@ -35,8 +35,8 @@ import {
 import Time from '@stackstorm/module-time';
 
 @connect((state) => {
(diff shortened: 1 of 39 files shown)


enykeev:
Right now, I don't see a way to get an intermediate execution state other than listening to all the executions progress all the time. Browser can't read raw sockets so we either need to start outputting eventsource-formatted messages with appropriate header or create another endpoint that outputs execution progress up to this point and properly closes the connection.

/cc @kami

LindsayHill:
Have been doing some quick testing with this.

Some initial feedback:

- Firefox not happy - high CPU. Not sure what's causing it yet.
- Sometimes very slow loading, but this may be related to my Docker setup. Suspect more server-side problem than client-side. As example, History is slow to load - first click shows no execution history, then a few seconds later it loads some content.
- When viewing output on the code app, there is no visual indication if "Wrap lines"/"Show newlines" is selected. Clicking them **does** change the way the output is displayed, but does not change the button color as it used to.
- I don't like the way the "Wrap lines"/"Show newlines" buttons are at bottom right, and the "Back" button is at top left. I also feel they should have more similar visual styling, but I will defer to UX experts on this matter.
- If I have a long-running action, when I view the history tab, Under "Action Output", it says "// Action produced no data". If I follow the "check live output" link, I see data (as expected). Maybe we need to change "Action produced no data" to "Action currently executing...click <here> to see live output"

PR #599, opened 2018-07-27
Inquiry app

enykeev:


nmaludy:
+1 will this be bundled in the unsable branch on PackageCloud?

LindsayHill:
@shusugmt I think you will want to take a look at this one

LindsayHill:
@nmaludy you can pull the RPM/Deb artifacts from https://circleci.com/gh/StackStorm/st2web/1042#artifacts/containers/0

nmaludy:
I can confirm this works, got a list of my inquiries.

Some comments:
- Unable to respond to an inquiry
- Would be nice to see information / link to the execution that triggered the inquiry

enykeev:
I should have probably pointed out that it is an early WIP. I'm just going breath first over the improvements for 2.9 to figure out what's missing on an st2 side.

In this particular case, it seems to me that we need more information on the inquiry to make UI not look empty.

@nmaludy you seem to be using inquiries day to day. What kind of information would you like to see displayed in the Inquiry app outside of the fill-out form and a link to execution (which by the way seem to share the same ID with the inquiry itself)?

enykeev:
I've decided to switch back to using `/executions` endpoint to get inquiries so I have a wider set of data to work with without switching context. My intuition tells me that we should release inquiry app this way, make the same transformations in inquiriy endpoints at the later point in time and only after that move them from experimental to v1.

LindsayHill:
Comments on using this with st2web2.9-dev10:

- Showing 'succeeded' inquiries is cool. Makes me think maybe we also need the ability to explicitly reject an inquiry (obviously this needs lower-level support)
- As mentioned, need to see some context around the originating inquiry - what workflow is this, who raised it? Cross-link to item in history
- For both 'pending' and 'succeeded', would be good to see additional data such as timestamp. When was it raised, when was it approved/rejected?
- Notifications - maybe some sort of icon over "Inquiries" app, to show that there are pending inquiries? 
- Don't allow responding to previously responded to inquiries. Currently allows it, then throws an error.

PR #598, opened 2018-07-27
Request only essential fields from API

enykeev:

PR #605, opened 2018-09-11
Changes for 3.0.0 release

Kami:
Included changes:

* Request only essential fields from API - https://github.com/StackStorm/st2web/pull/598


Kami:
Synced it with latest master and tests are passing now.

I also confirmed all the tabs load. I will go ahead and merge it into master.

On a related note - I don't know why we need this - d97b60dc6a7778757f1bd5d8d557d9379195b541. Even if I manually specify all the attributes, I still see `TypeError: Cannot read property 'properties' of undefined` error in the console when loading Rules tab.

Looking at the code, it looks like something to do with payload_schema or parameters_schema, but those attributes are included so no idea why this error is throw if include_attributes is used.

PR #611, opened 2018-09-29
Switch to workflows for deploy step

LindsayHill:
Switch to workflows, with separate jobs for 'build' and 'deploy', to allow filtering of the deploy step based upon branch.

This is the recommended way of doing it

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -126,18 +126,52 @@ jobs:
       - store_artifacts:
           path: ~/artifacts
           destination: packages
-      - deploy:
-          name: Maybe Deploy
+      - persist_to_workspace:
+          root: /home/circleci/artifacts
+          paths:
+            - trusty
+            - xenial
+            - el6
+            - el7
+  deploy:
+    docker:
+      - image: ruby:2.4.4
+    environment:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -126,18 +126,52 @@ jobs:
...
           destination: packages
-      - deploy:
-          name: Maybe Deploy
+      - persist_to_workspace:
+          root: /home/circleci/artifacts
+          paths:
+            - trusty
+            - xenial
+            - el6
+            - el7
```
armab: Is it possible to simplify but just using `.` instead of hardcoding every path?
I guess there is nothing else in artifacts dir and so additional filtering is not needed.

LindsayHill: Looks like artifacts dir also contains logs:

image

Probably not all that big a deal if we also persist those to workspace. Would need to check what st2flow and bwc-ui repos are doing here too, since they follow the same pattern.

LindsayHill: Will leave this as current pattern for now, if we change it to be less specific, we will do it as change across all repos where we use this design pattern

PR #617, opened 2018-10-05
Fixes for st2flow integration

DesignByOnyx:
This is a set of changes that allows st2flow to share functionality with st2web. As of right now we are needing to share the `st2-styles` and `st2-auto-form` packages.

NOTE: this PR is tightly coupled to [this PR on st2flow](https://github.com/StackStorm/st2flow/pull/151).

This PR also contains minor changes for the following:
- fix tests for `time` component so they work in regions with daylight savings
- allow `--reporter` flag to be set while running gulp tests


bigmstone:
Closing. Seems to be stale. We can re-open if needed.

Issue #651, opened 2018-12-13
Login font uses serif font


blag:
The login screen uses a serif font, but the fonts used in the rest of st2web are all sans serif fonts. This makes the login screen look a little weird:

<img width="272" alt="screen shot 2018-12-13 at 3 21 20 pm" src="https://user-images.githubusercontent.com/597113/49973724-12253c80-feeb-11e8-8eda-49555cda2896.png">

m4dcoder:
Fixed at https://github.com/StackStorm/st2web/pull/656

PR #656, opened 2019-01-09
Issue 651 login font uses serif font #651  

VineeshJain:


modules/st2-login/style.css
```diff
@@ -104,7 +105,7 @@
 
   &-label {
     font-weight: 100;
-
+	font-family: Roboto, sans-serif;
```
armab: Nitpick: would be good to use original separator as spaces, not the tabs for consistency with existing codebase

VineeshJain: Fixed.

PR #696, opened 2019-04-18
WebUI does not auto refresh with clean browser

### LINKED ISSUE #695: Web UI does not refresh with a clean browser cache
UI does not auto-refresh if any actions are run from outside. This only happens if the UI is launched once the browser cache/cookies are all cleaned up. 

Here are the steps to reproduce the issue:

1/ Open Firefox, with a blank page. Clear all history, cache, everything. Restart Firefox (current GA version)
2/ Go to https://<st2_ip>/, login.
3/ Open SSH session. Run `st2 run core.local cmd=date`. Run it a few times.
4/ Watch history tab in background. Observe that it does not update with recent executions.
5/ Go to Firefox. Refresh the page. See that recent executions are now displayed
6/ Go back to CLI. Run `st2 run core.local cmd=date` again.
7/ See that Firefox is now updating to show those new executions without needing to reload the page

image

Check out above. Firefox on the left, Chrome in the middle, terminal on the right.

When I run the action via CLI, it promptly shows up on Chrome. It does *not* show up on Firefox.

But the key thing here is that if I manually refresh the tab in firefox, it then starts auto-refreshing

VineeshJain:
With a clean Browser cache/cookies, when the UI was launched, the
call to /stream used to fail due to which the UI did not refresh.
This PR fixes the issue by sending the auth-token in the /stream
request. This PR also updates the version of Eventsource library to
the latest version (1.0.7) which allows headers to be set in the
Eventsource request.

Closes #695


modules/st2-api/api.js
```diff
@@ -255,7 +256,8 @@ export class API {
     return new Promise((resolve, reject) => {
       try {
         const source = new EventSource(streamUrl, {
-          rejectUnauthorized: this.rejectUnauthorized,
+          https: {rejectUnauthorized: this.rejectUnauthorized},
+          headers: {'X-Auth-Token': this.token.token},
```
armab: :100:

That's a nice assembled PR! :+1: 
I bet it took quite of time to debug and catch it .

---

FYI, some history that could be related or at least good to know: https://github.com/StackStorm/st2/pull/3566

VineeshJain: Thanks Eugen. Yes it took a while to get a handle on the issue. Now it is taking long time to figure out the CircleCI tests failing

Kami:
I was also able to reproduce this issue the other day, so good work :+1:

armab:
@VineeshJain Can you please finish this PR so we can merge it? Looks like there are some tests failing.

Issue #699, opened 2019-04-23
"Save" button not visible for pack configs on Firefox


LindsayHill:
Issue seen with both 2.10.4 and 3.0-dev6. 

On Firefox 66.0.3, the pack configuration section on the packs UI tab looks like this:

![image](https://user-images.githubusercontent.com/3607046/56607567-ae5c0000-65bd-11e9-8f05-e37e72dd8b7b.png)

Note that "Preview" is cut off, and "Save" is not visible. There is no horizontal scroll available to get to "Save". The only option is to 'tab' through the buttons until it displays like this:

![image](https://user-images.githubusercontent.com/3607046/56607597-ca5fa180-65bd-11e9-90cd-87d0a1f58161.png)

Note it is now cutting off part of the left-hand side of this panel.

This is how it renders on current GA Chrome versions:

![image](https://user-images.githubusercontent.com/3607046/56607630-dc414480-65bd-11e9-87e1-f5f1f46e5247.png)

That is what I would expect to see on Firefox. 

Safari looks the same as Firefox.

PR #702, opened 2019-04-29
Save Button is not visible in Packs Config on Fireox

VineeshJain:
The .st2-panel--detailed .st2-details class used by the Panels
packs-details component had display: table set though the contents
were not of type table. Removing the display:table fixed the issue


VineeshJain:
Fixes #699

VineeshJain:
Thanks Lindsay

VineeshJain:
Closes #699

Issue #710, opened 2019-05-14
Bug: Array of type object not validating 


jdmeyer3:
If an action has the following for a parameter type in an action:
```yaml
parameters:
  test_array_obj:
    type: array
    items:
      type: object
    description: "array of objects to evaluate"
    required: true
    position: 0
```
There is an issue with the UI where I cannot enter an array of objects into the parameter field.
![image](https://user-images.githubusercontent.com/25443446/57704218-b6d4b300-761e-11e9-94c3-9f81b58cba59.png)

I've tried many permutations on how to enter the array of objects, but the UI cannot evaluate it correctly.

## Installed Versions
st2.x86_64                         2.10.1-1               @StackStorm_stable
st2-auth-ldap.x86_64               2.10.4-1               @StackStorm_enterprise
st2chatops.x86_64                  2.10.1-1               @StackStorm_stable
st2flow.x86_64                     2.10.4-1               @StackStorm_enterprise
st2mistral.x86_64                  2.10.1-1               @StackStorm_stable

## Distro 
CentOS 7

MichaelMcClure:
This might be relevant:  https://github.com/StackStorm/st2web/issues/394

LindsayHill:
Resolved by #711

VineeshJain:
Reopening the issue as the proposed fix caused breakage in array type fields. Please refer to #711 for further details.

PR #711, opened 2019-05-15
issue-710 fixed array of objects not validating

jdmeyer3:
This PR deals with #710 where if a form has the datatype "array of objects" it would never validate it. It also deals with the string mutation of an array of objects, causing any entered array of objects to be mutated to [object Object].

### DIFF
--- a/modules/st2-auto-form/fields/array.js
+++ b/modules/st2-auto-form/fields/array.js
@@ -17,9 +17,9 @@ const typeChecks = (type, value) => {
   const v = String(value);
   switch (type) {
     case 'number':
-      return !validator.isFloat(value) && `'${v}' is not a number`;
+      return !validator.isFloat(v) && `'${v}' is not a number`;
     case 'integer':
-      return !validator.isInt(value) && `'${v}' is not an integer`;
+      return !validator.isInt(v) && `'${v}' is not an integer`;
     case 'object':
       return !_.isPlainObject(value) && `'${v}' is not an object`;
     case 'string':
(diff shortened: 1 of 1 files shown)


jdmeyer3:
I would like to give a special thanks to @humoroushorse for helping me find the exact issue.

Kami:
Thanks for the contribution.

@bigmstone / @VineeshJain Any chance you will have some time to review this?

jdmeyer3:
@bigmstone / @VineeshJain Have you guys had any chance to review this PR?

VineeshJain:
**@jdmeyer3** Reverted the code. This fix breaks the basic array functionality of a simple array type field. Any of the actions with type:array failed to load. Some of the sample actions that were breaking are:

`./examples/actions/orquesta-with-items-concurrency.yaml`
`./examples/actions/orquesta-test-with-items.yaml`
`./examples/actions/mistral-repeat-with-items.yaml`
`./examples/actions/orquesta-with-items.yaml`
`./examples/actions/mistral-jinja-repeat-with-items.yaml`

 Can you please take a relook at this? From the quick look at things, I found that the above actions don't have the `spec.items` field defined and thus using `spec.items.type.type` in the if condition causes errors.

modules/st2-auto-form/fields/array.js
```diff
@@ -71,14 +71,18 @@ export default class ArrayField extends BaseTextField {
...
-    if (jsonCheck(v) || Array.isArray(v)) {
+    if (jsonCheck(v)) {
       return JSON.stringify(v);
     }
 
     if (isJinja(v)) {
       return v;
     }
 
+    if (Array.isArray(v) && this.props.spec.items.type === 'object') {
```
VineeshJain: This is the offending line in the scenario where actions may not above spec.items defined

jdmeyer3: You guys are right, I did not test with the yaml definition not having the type of item. Let me take another look on how I can distinguish what type of items are in the array.

PR #748, opened 2020-04-27
Remove testing and deployment for Ubuntu Trusty

blag:
The `PKG_RELEASE` assignment is (I think) used to determine what the next release version string should be. As we deprecate OSes, we will need to update that to rely on a continuous stream. If I'm correct, it should not be set to an OS that we have not yet deployed on yet. I bumped it to use the Ubuntu Xenial packages, since those have a long rich release history.


blag:
This will also need to be merged into the `v3.2` branch.

PR #754, opened 2020-05-13
Stackstorm update gulp-csscomb

guzzijones:
updated gulp-csscomb in order to fix deprecated minimatch


guzzijones:
@bgaeddert yes I had planned on attacking each gulp dependency individually.

PR #757, opened 2020-05-28
Dependency clean up

bgaeddert:
The goal of this PR is the get a handle on the dependencies of st2web. The main purpose was to mitigate or remove all the packages with security vulnerabilities. While attempting this I found a few other issues that needed to be addressed first. First I went through all the modules and apps with package.json files and removed all unused dependencies. Then moved all devDependencies and production dependencies to their appropriate group. This process had the benefit of removing several vulnerabilities that weren't even used. And I was able to isolate the remaining vulnerabilities to devDependencies group. Then actually address the production packages vulnerabilities. The command below return vulnerabilities for production code. And as you can see there are now zero vulnerabilities.

$ `yarn audit --groups dependencies`

yarn audit v1.22.4
0 vulnerabilities found - Packages audited: 120
✨  Done in 2.24s.

There are remaining vulnerabilities in the devDependecies. However, IMHO it's not worth replacing packages that are working now to build and dev. I suspect that Githubs automated security check will still alert on these non-producton packages, but that remains to be seen.

I ran test and build and tested with st2vagrant. However, Some real QA is required and someone who knows something about the production build process should also test.

@punkrokk please assign anyone you feel appropriate.


LindsayHill:
Nice work here. Just removing unused dependencies alone will remove a whole lot of noise in future

nmaludy:
Tested locally on my box and didn't run into any issues. Works great!

bgaeddert:
@punkrokk @blag Is this ready to merge?

punkrokk:
@bgaeddert Yup! Thanks man. Great job!

Issue #758, opened 2020-05-28
Integrate st2flow (Workflow Designer) into st2web


armab:
EWC (previously StackStorm Enterprise) was opensourced with Workflow Designer (https://github.com/stackstorm/st2flow) is now free & open.

Integrate st2flow into st2web, as it previously was with the Enterprise Web UI.

PR #759, opened 2020-05-28
Integrate st2flow

mickmcgrath13:
Closes https://github.com/StackStorm/st2web/issues/758

This just copies the relevant modules from st2flow to st2web and changes a few things to wire it up to work.

It adds a "Workflows" top nav item:
image

It also adds a new route for `/action` which has the standard ST2 UI Header:
image

It adds an "Edit" link in the details panel of actions that are orquesta or mistral workflows:
image

It adds a "new" link (plus button) to the bottom right of the actions list on the actions page:
image

# how to test
```
git clone git@github.com:StackStorm/st2web.git
cd st2web
git checkout integrate-st2flow
```

add a file called config.local.js like:
```
angular.module('main')
  .constant('st2Config', {

    hosts: [
      {
        name: 'Verizon Training',
        url: 'https://st2instance.com/api',
        auth: 'https://st2instance.com/auth',
        stream: 'https://st2instance.com/stream',
      },
    ],

  });
```

and then run
```
docker-compose up
```

# Known issues:

## eslintrc
~see `notice/notice` in  `.eslintrc.yml`~ Resolved

## Super expression error 
(partially resolved.. see comments)

## Rebase with master
Master has updated to clean up old and unused dependencies (among other things).

I've attempted this locally, and changes to the files (even with a `docker-compose up --build` and forcibly pruning docker images) didn't seem to have an effect on the built files.

> **Update:** This is done

## CSS
I didn't copy any of the css or `modules/st2-style` from st2flow to st2web

Some icons are missing.  I didn't get a chance to test anything else

# Remaining items

- [x] eslintrc.yml - notice/notice (as described above)
- [x] Rebase/merge with master
- [ ] Clean up st2flow packages old / unused dependencies
- [ ] CSS (as described above)
- [ ] Testing (create a new workflow, edit a workflow, etc)
- [ ] Migrate the relevant issues from st2flow


mickmcgrath13:
re Super expression error:
- [st2flow](https://github.com/StackStorm/st2flow/blob/master/package.json#L71): "react-redux": "^6.0.1",
- [st2web](https://github.com/StackStorm/st2web/blob/master/package.json#L80): "react-redux": "7.0.2",


based on feedback from Bitovi internal, I changed this
```
// child
@connect(
  ({ ranges, notifications }) => ({ ranges, notifications })
)
export default class Editor extends Component<{
//parent
class MetaEditor extends Editor {}
```

to this:
```
// child
const editorConnect = ({ ranges, notifications }) => ({ ranges, notifications });
export { editorConnect };
export default class Editor extends Component<{
// parent
import { editorConnect } from '@stackstorm/st2flow-editor';
@connect(
  editorConnect
)
class MetaEditor extends Editor {}
```

mickmcgrath13:
based on feedback from Bitovi internal, I changed this
```
// child
@connect(
  ({ ranges, notifications }) => ({ ranges, notifications })
)
export default class Editor extends Component<{
//parent
class MetaEditor extends Editor {}
```

to this:
```
// child
const editorConnect = ({ ranges, notifications }) => ({ ranges, notifications });
export { editorConnect };
export default class Editor extends Component<{
// parent
import { editorConnect } from '@stackstorm/st2flow-editor';
@connect(
  editorConnect
)
class MetaEditor extends Editor {}
```


for better or worse, this helped...  will evaluate if this is a proper solution

mickmcgrath13:
re
> will evaluate if this is a proper solution


It is a serviceable solution.  See discussion on Bitovi's community slack:  [here](https://bitovi-community.slack.com/archives/CFD3N8N6T/p1590686750004700)

_( if you don't have access to Bitovi's community slack, sign up here: https://www.bitovi.com/community/slack) _

modules/st2flow-canvas/collapse-button.js
```diff
@@ -0,0 +1,58 @@
+// Copyright 2020 Extreme Networks, Inc.
+//
+// Unauthorized copying of this file, via any medium is strictly
+// prohibited. Proprietary and confidential. See the LICENSE file
```
armab, line 4: We can't just copy these files with the old proprietary license.
We'll need to first change the license to Apache 2.0 and remove Extreme Networks EULA-related things.

I created dedicated issue for that https://github.com/StackStorm/st2flow/issues/374 to track that as part of the transition.

mickmcgrath13: Should we even use st2flow anymore?  This PR really makes that separated repo unnecessary because all of the relevant components from st2flow were literally copied into this repo (via this PR).

Is there a legal reason we need the other repo?

cc @armab @m4dcoder

armab: Sure, we can archive `st2flow` repo once the functionality gets integrated into `st2web` in full.
However we'll need to update the license from the Extreme proprietary to OSS Apache 2.0 in any way.

mickmcgrath13: I agree.  commented as such here:
https://github.com/StackStorm/st2flow/issues/374#issuecomment-638873860

m4dcoder:
@mickmcgrath13, I came across the following issue when trying out this PR. I am not using docker compose to run it.  I'm using `make build` and `make install` and then run st2web from nginx.

When I edit a workflow or open a blank canvas with the "plus" button, a new tab open but it is blank and I get the following error in the developer tools on Chrome.

```
main.js:1 Invariant Violation: Could not find "store" in the context of "Connect(Workflows)". Either wrap the root component in a <Provider>, or pass a custom React context provider to <Provider> and the corresponding React context consumer to Connect(Workflows) in connect options.
    at module.exports (https://172.168.50.28/js/main.js:1:2405836)
    at Connect.renderWrappedComponent (https://172.168.50.28/js/main.js:1:674543)
    at Connect.indirectRenderWrappedComponent (https://172.168.50.28/js/main.js:1:674421)
    at Tg (https://172.168.50.28/js/main.js:1:3055949)
    at bi (https://172.168.50.28/js/main.js:1:3077684)
    at ci (https://172.168.50.28/js/main.js:1:3078058)
    at Di (https://172.168.50.28/js/main.js:1:3084899)
    at Yh (https://172.168.50.28/js/main.js:1:3084286)
    at Xh (https://172.168.50.28/js/main.js:1:3083323)
    at qf (https://172.168.50.28/js/main.js:1:3082208)
```

m4dcoder:
I'm still getting the same error. When I hit the plus button, a new tab is created but then it's blank. At the developer console, I see the following error.  I'm not running this in Docker.  I run make build and install and then configured it using nginx locally. This is similar setup to st2web package.  Have we tried whether this is working using st2web dev packages? 

```
Invariant Violation: Could not find "store" in the context of "Connect(Workflows)". Either wrap the root component in a <Provider>, or pass a custom React context provider to <Provider> and the corresponding React context consumer to Connect(Workflows) in connect options.
    at module.exports (https://IP_ADDRESS/js/main.js:1:2348623)
    at Connect.renderWrappedComponent (https://IP_ADDRESS/js/main.js:1:656683)
    at Connect.indirectRenderWrappedComponent (https://IP_ADRESS/js/main.js:1:656561)
```

mickmcgrath13:
## Build
> I run make build and install and then configured it using nginx locally.

this is running [make build-and-install](https://github.com/StackStorm/st2web/pull/759/files#diff-3254677a7917c6c01f55212f86c57fbfR11), too:
```
build-and-install:
	make build
	make install
```

The [Readme says to run](https://github.com/StackStorm/st2web/tree/integrate-st2flow#production)
```
gulp production
```
which ultimately is what [make build](https://github.com/StackStorm/st2web/blob/integrate-st2flow/Makefile#L38) does via [npm run build](https://github.com/StackStorm/st2web/blob/integrate-st2flow/package.json#L9):
```
"build": "gulp production",
```


## Serve
The Dockerfile is [starting up](https://github.com/StackStorm/st2web/pull/759/files#diff-3254677a7917c6c01f55212f86c57fbfR17) with `npm run serve`
..[which in turn runs](https://github.com/StackStorm/st2web/pull/759/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R10):
```
gulp serve-production
```

How is yours being started? just nginx? no node?
What does the nginx config look like?

IMO, It should work through docker, else the `Dockerfile` isn't relevant (don't we need it for kubernetes based deployments anyway?).

How is the Dockerfile being used today?

I'm not sure what the difference is between what happens with `npm run serve` and an nginx-served site, but if we want it served with NGINX, we should update the Dockerfile to start up and serve via NGINX rather than node so that it's consistent.


> This is similar setup to st2web package

I'm not sure what this means.  Isn't this [actually the st2web package](https://github.com/StackStorm/st2web/blob/integrate-st2flow/package.json#L2)?

> Have we tried whether this is working using st2web dev packages?

I'm not sure what this means.  Can you provide a link to some docs or clarify what this means?

mickmcgrath13:
[gulp serve-production seems to be serving `./build`](https://github.com/StackStorm/st2web/blob/integrate-st2flow/tasks/production/serve.js#L24):
```
gulp.task('serve-production', gulp.series([ 'production' ], () => {
  server = gulp.src('./build')
    .pipe(plugins.webserver({
      host: '0.0.0.0',
      port: 3000,
    }));

  return server;
}));
```

[make install](https://github.com/StackStorm/st2web/blob/integrate-st2flow/Makefile#L44) seems to be copying things **from** build/ to a destdir:
```
install:
	echo "install"
	mkdir -p $(DESTDIR)$(PREFIX)
	cp -R $(CURDIR)/build/* $(DESTDIR)$(PREFIX)
```

I wonder if `gulp serve-production` is still looking at the `$(CURDIR)/build/*` directory when it should be looking at `$(DESTDIR)$(PREFIX)`.

Where does `$(DESTDIR)` and `$(CURDIR)` get set to be used in the Makefile?  Is that a Makefile thing?  I only see [PREFIX](https://github.com/StackStorm/st2web/blob/integrate-st2flow/Makefile#L4):
```
PREFIX ?= /opt/stackstorm/static/webui
```

m4dcoder:
@mickmcgrath13 Here's the basic single node st2 installer where is setting up nginx and st2web.  https://github.com/StackStorm/st2-packages/blob/master/scripts/st2bootstrap-deb.sh#L589. I'm just pointing out that I'm not able to get st2flow to work in my setup, which follows this install script. I think the best thing to do is to get CI to pass here and have the st2web packages built so we can verify it thru normal installation means.

armab (approved):
This is so awesome, congrats! :tada: 
So can we merge it?

@mickmcgrath13 @m4dcoder Did you guys just fix all the failing tests here and so we could release the Flow in `v3.3.0`?!

amanda11:
@mickmcgrath13 
I failed to get the new Workflows tab or edit workflow working on my CentOS7 system. I installed the new st2web el8 from the CircleCI run, and did a restart of all st2 services with st2ctl restart.
I'm seeing following in console if I try and edit a worflow:

image

armab (changes requested):
Thanks @m4dcoder @amanda11, I can confirm I'm getting same error message when installing the st2web via `deb/rpm` packages produced during the build in this PR.

I understand now that's the main blocker.

m4dcoder (changes requested):
Additional changes required. Fonts are off in the st2flow UI. Zoom icons in the st2flow UI are missing. The workflow icon in the main st2web UI upper nav menu does not open st2flow in new tab whereas the plus sign opens st2flow in a new tab. We will remove the workflow icon until we figure out better UX. The focus is to keep UX same as EWC. Blocking here so this won't be merged for v3.3 until this PR/feature is ready.

amanda11:
Excellent to see it up and running, was able to import in an existing orquesta workflow. Can see community loving this feature.

Add/Edit/Delete actions worked well (just a minor problem I found on edit). Adding transitions ok, but had problems for edit/delete - see below.

Some observations I had on top of what @m4dcoder wrote:

1. If I am on the history tab and click the Action icon - then both the Action and Workflow button get highlighted in blue. If the workflow icon is being removed, then this might not be relevant
2. On actions that have boolean parameters, there is a "reset default" button, this didn't seem to reset for me (but it does on the old EWC). e.g. Take the chatops.post_message action, check the box next to whisper, then hit the "reset default", and it doesn't clear the checkbox (but it would have done with EWC)
3. If I click a transition - I get a blank screen generated. I'll attach the screenshot with the console error. So I can't edit or delete a transition.
image
Same problem if I click the Transitions tab.
4. I'm on chrome and toggles on the action/properties came up quite small. The colour changes, but I can't see a positional change - like I could on the sliders in EWC. Screenshot attached:
image

amanda11:
@m4dcoder Not sure what the intention is, but in the old EWC the + icon to create a new workflow, and edit on an existing workflow both opened the workflow designer in a new tab. In the latest version of this PR then edit opens the workflow designer in a new tab, but + doesn't open a new tab. 
Zoom icons worked well for me now (tested on Chrome and CentOS 8 - and all working great).
No additional comments to those I mentioned in my previous comment above.

m4dcoder:
@amanda11 My intention is to not open the workflow designer in a new tab. We'll revisit the edit to make sure the behavior is consistent.

m4dcoder:
Shital, Here are a few more issues to address on the latest here, priority in the order listed. Thanks!

1. When I open an existing workflow examples.orquesta-sequential, the transition between tasks (the lines connecting the tasks) does not show up until I click on the canvas.
2. When I open an existing workflow examples.orquesta-sequential and then click on the task transition (the line connecting the tasks), the whole app goes blank and I cannot exit from there until I close the tab and log back into st2web UI.
3. When I open an existing workflow in st2flow UI (say any examples.orquesta-*), the first time I open the workflow, the st2flow UI always expect me to save the workflow before I can execute it. If I didn't make any changes to the workflow, the st2flow UI shouldn't prompt me to save it.
4. Let's say I open an existing workflow in st2flow UI, executed the workflow, and then go back to the history tab. If I click on the "+" button, the st2flow UI still has my first workflow open. Can we reset the st2flow UI when the "+" button is used to go into st2flow UI?
5. Let's say I open the workflow examples.orquesta-sequential in st2flow UI, if I execute this workflow, the execute workflow button is grayed out until I provide a value for "name" even though there is a default value already there.  The execute workflow button should not be grayed out if there's default value provided.
6. In the st2flow UI, the "x" button next to any checkboxes on the right hand panel doesn't do anything.

amanda11 (changes requested):
Looks good, the issues I had found before are all fixed, and it is good.

I found one problem that is new to this that wasn't in the EWC version.

If I click an action in a workfow, and then click  Properties tab. If I slide the Join to on, then it shows me Join All Tasks or Join n tasks - but there is no check box to show which option is chosen.

JoinCheckBoxMissing

On the EWC version I would have seen checkboxes:
JoinEWC

amanda11:
> Looks good, the issues I had found before are all fixed, and it is good.
> 
> I found one problem that is new to this that wasn't in the EWC version.
> 
> If I click an action in a workfow, and then click Properties tab. If I slide the Join to on, then it shows me Join All Tasks or Join n tasks - but there is no check box to show which option is chosen.
> 
> JoinCheckBoxMissing
> 
> On the EWC version I would have seen checkboxes:
> JoinEWC

Verified this problem is now resolved.

amanda11 (changes requested):
I found a couple of problems that exists in this version of st2web that didn't exist in EWC (though that may be the same underlying problem). The scenarios are:

Scenario 1 - Loose Save indicator on Re-arrange workflow
1. Add an action to a saved workflow - as there is unsaved changes the Save icon has a red outline to indicate there are unsaved changes
2. Hit the re-arrange tasks button. The Save icon looses it's red outline - so you don't know there are unsaved changes.

In EWC then the button would have remained red.

Scenario 2- moving a task doesn't indicate Save is needed
1. Edit an existing workflow and move the locaiton of a task. The Save indicator doesn't go red to indicate a change has been made. In EWC moving a task would have showed that a change had been made.

amanda11 (approved):
Workflow composer looks great, had a good play with editing and creating new workflows. Lot of testing on the Save indicator, and that all looks good.
Excellent work.

m4dcoder (approved):
Shital, This is an awesome job! Thank you for brining this effort to completion.

armab (changes requested):
Per initial @mickmcgrath13 message, PR added the `Workflow` tab nav item and I found that brilliant idea as previously the `st2flow` feature was pretty much hidden from users.



I couldn't find `Workflows` button in current implementation. Please include that too.

amanda11:
I believe the workflow tab caused some problems, and so was taken out of this PR a few rounds back. I think the intention was to get equivalent functionality to EWC in and then try and get the workflow tab in separately.

amanda11:
> Additional changes required. Fonts are off in the st2flow UI. Zoom icons in the st2flow UI are missing. The workflow icon in the main st2web UI upper nav menu does not open st2flow in new tab whereas the plus sign opens st2flow in a new tab. We will remove the workflow icon until we figure out better UX. The focus is to keep UX same as EWC. Blocking here so this won't be merged for v3.3 until this PR/feature is ready.

@armab This is the comment from a few months ago when the workflow tab was taken out. (I know it was definitely inconsistent and getting a bit confused with the Action icon as well at the time).

armab:
@amanda11 Thanks!

@m4dcoder Was the issue with the `Workflows` button in nav menu only about opening Workflow Designer in new window vs current window? If that's the only reason, would be great to keep that change, - doesn't matter new/existing window as it's highly helpful addition for the users.

amanda11:
Had another re-test on latest version - looks good to me.

blag:
From the TSC: We need to update copyright headers in all repo files in this PR.

apps/st2-workflows/index.js
```diff
@@ -0,0 +1,24 @@
+// Copyright 2019 Extreme Networks, Inc.
+//
```
armab, line 2: @m4dcoder Here is an example with the Copyright headers update we need for every file that includes Extreme Networks (c): 

```suggestion
// Copyright 2020 The StackStorm Authors.
// Copyright 2019 Extreme Networks, Inc.
//
```

See https://github.com/StackStorm/st2/pull/5003 for the `stackstorm/st2` analogy.

m4dcoder (approved):

apps/st2-workflows/index.js
m4dcoder: Done

armab (approved):
Thanks everyone involved in this! :tada:

blag (approved):

PR #798, opened 2020-09-03
Workflow integration and functional and unit test cases execution issue fixes

shital-orchestral:
This patch fixes the issue reported at https://github.com/StackStorm/st2web/pull/759#issuecomment-675806306 on the integrate-st2flow feature branch the st2flow canvas is displaying an empty page with the following error. 

```
Invariant Violation: Could not find "store" in the context of "Connect(Workflows)". Either wrap the root component in a <Provider>, or pass a custom React context provider to <Provider> and the corresponding React context consumer to Connect(Workflows) in connect options.
    at module.exports (https://IP_ADDRESS/js/main.js:1:2348623)
    at Connect.renderWrappedComponent (https://IP_ADDRESS/js/main.js:1:656683)
    at Connect.indirectRenderWrappedComponent (https://IP_ADRESS/js/main.js:1:656561)
```

This patch removes the typescript code from the workflow.component.js and wrap all that content inside <Provider> tag in workflow.componet.js.

### DIFF
--- a/apps/st2-workflows/workflows.component.js
+++ b/apps/st2-workflows/workflows.component.js
@@ -14,26 +14,20 @@
 
 import React, { Component } from 'react';
 // import ReactDOM from 'react-dom';
-// import { Provider, connect } from 'react-redux';
-import { connect } from 'react-redux';
+import { Provider, connect } from 'react-redux';
 import { PropTypes } from 'prop-types';
 import { HotKeys } from 'react-hotkeys';
 import { pick, mapValues, get } from 'lodash';
 import cx from 'classnames';
 import url from 'url';
-
 import Menu from '@stackstorm/module-menu';
 import Palette from '@stackstorm/st2flow-palette';
 import Canvas from '@stackstorm/st2flow-canvas';
 import Details from '@stackstorm/st2flow-details';
-
 import api from '@stackstorm/module-api';
-
-// import CollapseButton from '@stackstorm/st2flow-canvas/collapse-button';
 import { Toolbar, ToolbarButton, ToolbarDropdown } from '@stackstorm/st2flow-canvas/toolbar';
 import AutoForm from '@stackstorm/module-auto-form';
 import Button from '@stackstorm/module-forms/button.component';
-
 import { Route } from '@stackstorm/module-router';
 import globalStore from '@stackstorm/module-store';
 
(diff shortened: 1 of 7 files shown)


armab (commented):
Thanks for contribution.

Can you please highlight in more detail what are the workflow integration fixes provided in this PR?
Do we have Issues documented somewhere for these fixes?

shital-orchestral:
**Reference error : performance is not defined while executing functional test cases**
1) These functional test cases are written to run on server. But inside st2flow-perf\index.js , due to some invalid condition It was referring to browser's object which is **performance** object that why it was showing undefined error .

2) To resolve this issue  we have change below conditional statement code from st2flow-perf\index.js
**const IS_NODE = typeof process === 'object' && Object.prototype.toString.call(process) === '[object Object]';**
3)After performing above action functional test cases got execute but with fail status

4) To pass all these test cases need to do following changes
-  Most of the functional test cases were not getting passed because of login authentication. To get login test case  pass we required following environment variables -ST2_USERNAME/PASSWORD/HOST  and it should get assign to proper value.
-The circle CI job should have these environment variables ST2_USERNAME/PASSWORD at https://github.com/StackStorm/st2web/blob/master/.circleci/config.yml#L84 before running the functional tests,maximum test cases will get passed after passing login test case.

5) Few code changes required inside **st2web\tests\test-login.js** and **st2web\tests\test-actions.js** to pass rest of test cases.
 -Need to do modification  **before hook of Submit function** inside **test-login** 
                      before(function () {
			this.timeout(20000);
			return util.login();
			});
-Need to do modification in below assertions inside **test-actions.js**
1) describe('Details view', () => {
    let resource;
     before(() => {
      resource = browser.resources.filter((e) => new      
    RegExp('**^https://example.com/api/v1/actions/views/overview/[\\w.-]+$**').test(e.url));
    });
     it('should make a call to actions endpoint', () => {
     // NOTE: expecting 3 calls because of filtering
    expect(resource).to.have.length.at.least(**1**,'Execution endpoint has not been called');
    expect(resource).to.have.length.at.most(**1,** 'Execution endpoint called several times');
    });
2)  describe('Details view', () => {
      let resource;
      before(() => {
  resource = browser.resources.filter((e) => new RegExp('^https://example.com/api/v1/actions/views/overview/core.announcement$').test(e.url));
      });
    it('should make a call to actions endpoint', () => {
        // NOTE: expecting 2 calls because of filtering
        expect(resource).to.have.length.at.least(**1**, 'Actions endpoint has not been called');
        expect(resource).to.have.length.at.most(**1**, 'Actions endpoint called several times');
      });

shital-orchestral:
**Reference error : performance is not defined while executing unit test cases**
1)To resolve this issue we have change below conditional statement code from `st2flow-perf\index.js`
**const IS_NODE = typeof process === 'object' && Object.prototype.toString.call(process) === '[object Object]' || Object.prototype.toString.call(process) === '[object process]';**
-**This change is required to run both unit as well as functional test case execution.** 
-After performing above action unit test case execution will get started but few assertions are getting failed,on **windows system** execution.
**2)json/yaml assertion check within various test case files was getting failed due to below reason**
-The assertions which are getting failed while matching two json/yaml file, in that case in our local **json/yaml** file end line characters are getting added like (\r\n). 
-Values/Content will be the  same in both files(json/yaml) but due to end line characters assertion gets failed.
-removed those characters by converting it into stringified data and replaced those characters  from data and then matched two files
e.g. I have made following code changes in one of the test case file assertion - \modules\st2flow-model\tests\test-model-meta.js
    it('writes meta-basic.json', () => {
  `   expect(JSON.stringify(model.toYAML())).to.equal((JSON.stringify(raw)).replace(/\\r/g, ''));`
    });
3)After performing similar changes in **st2flow-yaml\tests\test-token-set.js**, **st2flow-model\tests\test-model-orquesta.js** changes in these test files. All the test cases get passed.
 This  code modification is required to run test cases on windows but on linux system  it will work fine with/without the changes as I mentioned in second point .

mickmcgrath13 (approved):
The changes look good to me.  I haven't had a chance to actually test them, but I agree with the path and explanation.

Issue #811, opened 2020-10-13
WebUI: Add Donate button


armab:
Add `Donate` button in the visible place of the Web UI header linking to the https://stackstorm.com/donate/, similar to StackStorm website.

![linux dig_-_History_StackStorm_Web_UI_-_2020-10-13_14 20 37](https://user-images.githubusercontent.com/1533818/95866289-a5399280-0d5f-11eb-9c4f-c8f259a0ad15.png)

As project is struggling with the funding, this would help to make Donation button even more visible.

PR #812, opened 2020-10-13
Add donate button to st2web UI header

armab:
Closes https://github.com/StackStorm/st2web/issues/811

Packages installed from this PR CircleCI build shows the following "Donate" button linking to the https://stackstorm.com/donate/

### DIFF
--- a/modules/st2-menu/menu.component.js
+++ b/modules/st2-menu/menu.component.js
@@ -64,6 +64,7 @@ export default class Menu extends React.Component {
 
   docsLink = 'https://docs.stackstorm.com/'
   supportLink = 'https://forum.stackstorm.com/'
+  donateLink = 'https://stackstorm.com/donate/'
 
   handleDisconnect() {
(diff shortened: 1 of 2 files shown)


m4dcoder (changes requested):
I'm against adding the donate button here. This is distasteful. A little bit desperate.

armab:
OK, I'm adding this item up to discussion.

As we have `Donate` button on stackstorm.com, Github, I don't see how it's distasteful to have it on `st2web` too.
I never thought it would be something of disagreement as we're observing a problem supporting the project and covering its expenses (https://github.com/StackStorm/discussions/issues/36).

I'd better have it on st2web and allow broader Community supporting the StackStorm infra expenses, rather then see a day when users won't be able to install the tool when some system behind st2 infra goes offline as unpaid due.

cognifloyd:
I'm closing this as it is marked `wontfix`

PR #822, opened 2020-11-20
notification popup close on button click

shital-orchestral:
1)The issue was pop up on pack install don't go away after clicking show execution button.- This PR will fix this issue.It will close the notification popup on button click.


amanda11 (approved):
LGTM - looks good. Goes away when show execution clicked or you click the X or after the default timeout. So looks great.

PR #853, opened 2021-02-02
axios and urijs package version updated

shital-orchestral:
Updated Axios and urijs package versions.


shital-orchestral:
Created by mistake with  integrate-st2flow as a base branch

PR #854, opened 2021-02-02
Axios and urijs package version updated

shital-orchestral:
Axios package version updated to 0.21.1 and urijs to 1.19.4.
Thanks

### DIFF
--- a/apps/st2-workflows/package.json
+++ b/apps/st2-workflows/package.json
@@ -71,7 +71,7 @@
     "react-redux": "^6.0.1",
     "react-router-dom": "^5.0.0",
     "redux": "^4.0.1",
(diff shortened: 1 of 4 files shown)


blag:
@shital-orchestral Thank you for the contribution, but this breaks tests. Can you fix those?

armab (approved):
Nice one!

Thanks @shital-orchestral for the fix :+1:

m4dcoder (approved):
Great job!

armab:
@shital-orchestral @m4dcoder Looks like `yarn.lock` still includes the old version here:
https://github.com/StackStorm/st2web/blob/master/yarn.lock

Can you elaborate why it's still an old version and how that plays in the build?

PR #857, opened 2021-02-16
Update yarn.lock to match changes to urijs and axios updates

m4dcoder:
The yarn.lock file wasn't updated (requires make build) to reflect the
version changes to urijs and axios.


amanda11 (approved):
Looks good, and gave a quick test through with artificact.
All great. Thanks.

PR #858, opened 2021-02-17
Task delay model

shital-orchestral:
Added **Delay Property** as a new field under task properties, with the use of field, we can add delay in-between execution of the tasks. Thanks

### DIFF
--- a/modules/st2-auto-form/style.css
+++ b/modules/st2-auto-form/style.css
@@ -481,6 +481,19 @@
   }
 }
 
+.tooltiptext {
+  visibility: hidden;
+} 
+.showtooltiptext {
+   visibility: visible; 
+   width: auto;
+   padding: 0;
+   color: #6c6c6c;
+   background: none;
+   display: inline-block;
+   margin-top: 4px;
+   margin-left: 35px;
+} 
 
 
 
--- a/modules/st2flow-details/orquesta-properties.js
+++ b/modules/st2flow-details/orquesta-properties.js
@@ -46,91 +46,104 @@ type TransitionProps = {
   })
 )
 export default class OrquestaTransition extends Component<TransitionProps, {}> {
+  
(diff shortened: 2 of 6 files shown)


m4dcoder (changes requested):
@shital-orchestral 

1. The value of the task delay can be an expression (string) or an integer. Here is an example where the delay is an expression at https://github.com/StackStorm/st2/blob/master/contrib/examples/actions/workflows/orquesta-delay.yaml#L17. This current PR doesn't load this workflow definition and display the delay value correctly.
2. Is it possible to provide a hint when hover over the `delay` field which type of values user can input? For example, when hover over `delay`, hint that user can input an expression or integer.

m4dcoder (changes requested):
@shital-orchestral Can you make the delay input box the same as the delay in the task retry model? This will be more consistent.

m4dcoder (approved):

amanda11:
LGTM @m4dcoder @shital-orchestral  
The only feature I saw from testing was that I can enter -1 for delay, and the UI accepts this. But then when the workflow runs I get an error stating:
`TypeError: The value of task delay is not type of integer.",`

But I think this problem extends to other places as well, and its difficult to verify in the UI as we support strings for putting in contexts. So I think this is ok.

amanda11:
I don't seem to get a red box up when I'm in the field and enter -3 for delay.
I get one when I have the field as empty, but when it's a negative number then I don't see any message.

amanda11:
LGTM - I can see the negative number check now. Not sure if I had the wrong artificact before - but looks good.

m4dcoder:
@amanda11 Here you go. You should be seeing the error shown in the screenshot.

Screen Shot 2021-03-29 at 12 52 04 PM

amanda11 (approved):
LGTM

PR #863, opened 2021-03-07
fix bug: validation of array of objects

Moradf90:
fix the validation of the "array of objects", also the value of this when values (objects) separated by `,` (comma) #710


amanda11:
I think the array problems have been solved by https://github.com/StackStorm/st2web/pull/901

PR #871, opened 2021-04-02
Use st2-docker master docker compose setup for end to end tests

Kami:
WIP attempt to try to get end to end tests to use up to date st2-docker master branch.

Currently they use vastly outdated 1 year old deprecated ``stackstorm/stackstorm`` Docker image + compose setup.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -98,7 +98,7 @@ jobs:
       - run:
           name: Reset test containers
           command: |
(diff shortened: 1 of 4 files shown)


Kami:
OK, this was a pain, but it's all working and passing now :tada:

I will go ahead and merge it into my perf optimizations branch since that PR depends on changes from latest StackStorm/st2 master which means end to end tests will fail with 1 year old Docker image :D

PR #891, opened 2021-05-25
Update lodash version

amanda11:


amanda11:
Manuallly tested the st2web artifact built on EL8 using Chrome browser.

armab (approved):
Thanks!

PR #894, opened 2021-06-07
Update node.js to v14.16.1

amanda11:
Update node.js to v14.16.1. This is required to apply a security update that requires node.js to be v14+.


amanda11:
Tested package manually on single node install (both on a system using node 14 and one with node 10).

README.md
```diff
@@ -9,22 +9,22 @@ StackStorm Web UI
...
 ```shell
-$ n 10
+$ n 14
 
-install : node-v10.4.1
-  mkdir : /usr/local/n/versions/node/10.4.1
-  fetch : https://nodejs.org/dist/v10.4.1/node-v10.4.1-darwin-x64.tar.gz
+install : node-v14.17.0
+  mkdir : /usr/local/n/versions/node/14.17.0
+  fetch : https://nodejs.org/dist/v14.17.0/node-v14.17.0-darwin-x64.tar.gz
```
m4dcoder, line 19: How come the node version is different here?

m4dcoder (commented):
LGTM. I just have a minor question.

README.md
amanda11: They are different things...
1) In the README when you say n 14 - you get the latest version of node 14. The showing of the version here is just an example output - it will change each time they release a node version of n14.. 
2) In the Dockerfiles you are downloading the Node docker images
At the time of first creating this PR then the latest docker image for node 14, was 14.16. But when I took the output from my n14 command, it output 14.17.0 (now it will show a much larger version)

I could update these Dockerfiles - as I see now that Docker images for 14.17 are now available, but we don't generally update these with every patch of node.

However they will always end up getting out of synch, and if you look at the changes before this PR then we had for:
1) 10.4.1 in the README
2) 10.15.3 in the Dockerfile

m4dcoder: Ok. Thanks for the explanation.

m4dcoder (approved):

PR #901, opened 2021-07-20
Workflow composer displays blank if array field is a YAQL expression

### LINKED ISSUE #855: 3.4 Dev - st2flow Errors selecting Tasks
If the Packs.install workflow is loaded into ST2flow, clicking on certain tasks will cause the page to go blank and generate javascript errors in the Developer console.

On a brand new Centos8 install, clicking on the "stop_installation_and_cleanup_because_conflict" task as shown below:

image

Blanks the Flow GUI and generates the following errors in the Dev Console:

image

Reloading the page will cause correct the GUI display but clicking on the task again will repeatedly fail.

Clicking on other tasks will also generate the same error (always) but I have not been able to identify something common that would identify which task would fail.

amanda11:
When you have a workflow that calls an action that has an array parameter, but you pass in a yaql expression, you get a blank screen when you click on that action.

Update so that if the value passed to an arrayfield is a YAQL expression (ie. starts with <% and ends with %>) that it won't try and split on , and just return as is. i.e. the same logic it performs with Jinja expressions. As if a YAQL expression was found before then it would fail as it didn't have an array to split on.

Resolves #855


amanda11:
Verified with packs.install workflow that if select download_pack that would previously go to blank screen, that with this fix that it doesn't blank screen anymore.
Need to do more verification to check that it has stopped the ability to enter "value2","value3" and it convert to array etc.

amanda11:
Verified that behaviour on action input is unchanged, e.g.

- Go to actions packs.download - and enter `excel, aws_boto3` - and it still converts this to an array as it did before.

amanda11:
Performed some more validation, e.g. in Workflow Composer:

- added a task that required an array, and added value1, value2 on the input view - workflow composer still wrote this out as YAML as before (with each value on own line), and worked. And could edit that to add, delete.
- could add in workflow composer in YAML directly as ["value1","value2"] and it show correctly on the input view, and then when amend goes back to the value per line array.
- tested passing single values to array, and then get converted to array as  before
- tested passing in YAQL expression, and it just gets left alone, and therefore taken as an array.

BUT got a problem, in you have a string parameter such as param1, and then in the workflow definition need to pass an array. If you enter ["<%ctx().param1%>]" in the YAML directly then its all good, but it then gets shown in the input view without the []. But then if you then edit it, it will then add back in the [].

modules/st2-auto-form/fields/array.js
```diff
@@ -78,13 +78,26 @@ export default class ArrayField extends BaseTextField {
       return v;
     }
 
+    /* YAQL parameter that came input as single yaql */
+    if (isYaql(v) && !v.includes(',')) {
```
cognifloyd, line 82: What happens if there is a comma inside the YAQL expression?

cognifloyd (commented):
Looks like comma+YAQL might be problematic. Right?

modules/st2-auto-form/fields/array.js
amanda11: It will fall down to the split below. So it will go direct to the split below on the ,.

So solving "," in yaql is another issue, as if you have commas in your strings, or yaqls - it's going to go wrong at that split function.  

So I wasn't going to try and solve commas at this stage, as that will require a whole host more work. I think there's going to be edge cases that it doesn't split correctly, but trying to cope with simple YAQL as per the default ST2 pack installs etc, it should.

cognifloyd (approved):

m4dcoder (approved):
@amanda11 I requested @pradnya-orchestral to test and review this PR. She reviewed it and looks good. So approving this PR.

amanda11:
Thanks @m4dcoder.

PR #910, opened 2021-09-13
Allow integer fields with YAQL expressions to be saved

amanda11:
If we have an action that requires an integer or number as a parameter, and then we have a workflow that calls that action. Currently if we put a YAQL expression such as <% ctx().int_param %> as the value for the parameter, we will get an error stating that it is not a valid integer or float. 
Update the workflow composer so that it handles integer parameters and numbers with YAQL in the same way it handles Jinja.

Similar change made for objects. No change needed for booleans, but to put in YAQL or Jinja with booleans have to code to the YAML view.

### DIFF
--- a/apps/st2-actions/actions-details.component.js
+++ b/apps/st2-actions/actions-details.component.js
@@ -200,6 +200,18 @@ export default class ActionsDetails extends React.Component {
     return false;
   }
 
+  isExpression (value) {
+    if (value.startsWith('{{') && value.endsWith('}}')) {
+      return true;
+    } 
+    else if (value.startsWith('<%') && value.endsWith('%>')) {
+      return true;
+    } 
+    else {
+      return false;
+    }
+  }
+
   isValidInt (value) {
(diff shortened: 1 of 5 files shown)


pradnya-orchestral:
@amanda11 , @m4dcoder has asked me to check your changes, one thing here is, in action details page if integer field value is available ,we have added ranges for timeout and limit field and as per the ranges we are enabling and disabling run button. We have added condition for run button. I have tried your changes, now it is not giving error for YAQL expression but run button is disabled . So if user has entered YAQL expression for integer or number field we have to add appropriate condition for run button so we can enable that.

amanda11:
@pradnya-orchestral @m4dcoder Thanks for the review comments, I have updated the disabled to check whether its an expression and don't disable if it is. For expression checking I've added YAQL and Jinja, as the change to put the check on value, had stopped Jinja expressions such as {{ st2kv.system.int_param }} from being used (which could be used on other integers fields as a Run parameter).

m4dcoder (approved):

PR #950, opened 2022-01-01
Add link to trigger on History page

mingfang:
On the History page, there is a TRIGGER DETAILS section on the lower right.
The Trigger name is displayed but it would be better it linked to the actual Trigger page.
This PR turns the Trigger name into a link.


cognifloyd (commented):
I love the idea of lining to the trigger. I don't know all the JavaScript bits in st2web to decide if this is correct.

Could you add some before and after screenshots to highlight where this link will be added?

mingfang:
Before
image

After.  Note `git.head_sha_monitor` is now a link. It works the same was as the RULES DETAILS link above it.
image (1)

This is where it links to.  
image (2)

cognifloyd (approved):
LGTM

Issue #862, opened 2021-03-05
Error if save workflow that has a parameter in after an execute


amanda11:
I'm seeing some odd behaviour:

1. create a new workflow and give it a name
2. Give that workflow a parameter arg1 of type integer and default value of 3
3. Add a core.echo with message: Hello <% ctx().arg1 %>
4. Save
5. Run workflow
6. Hit save again...
I then see an error "Error saving workflow.: Additional properties are not allowed ('_name' was unexpected)"

I can reproduce it anytime I create a workflow with a parameter, hit save and then run, and then save. 
There may be other scenarios that trigger it.

The error comes back in a 400 HTTP response from the server.

But looking at the payload it does appear to be odd:
```
{"pack":"default","enabled":true,"runner_type":"orquesta","name":"test1","entry_point":"workflows/test1.yaml","parameters":{"arg1":{"type":"integer","default":"3","_name":"arg1"}},"metadata_file":"actions/test1.meta.yaml","data_files":[{"file_path":"workflows/test1.yaml","content":"version: 1.0\ntasks:\n  # [408, 180]\n  task1:\n    action: core.echo\n    input:\n      message: fdaf <% ctx().arg1 %>\ninput:\n  - arg1\n"},{"file_path":"test1.meta.yaml","content":"pack: default\nenabled: true\nrunner_type: orquesta\nname: test1\nentry_point: workflows/test1.yaml\nparameters:\n  arg1:\n    type: integer\n    default: 3\n"}]}
```

The value is _name on the save...

But on the initial create the message sent is:
```
{"pack":"default","enabled":true,"runner_type":"orquesta","name":"test1","entry_point":"workflows/test1.yaml","parameters":{"arg1":{"type":"integer","default":"3"}},"metadata_file":"actions/test1.meta.yaml","data_files":[{"file_path":"workflows/test1.yaml","content":"version: 1.0\ntasks:\n  # [408, 182]\n  task1:\n    action: core.echo\n    input:\n      message: fd33daf <% ctx().arg1 %>\ninput:\n  - arg1\n"},{"file_path":"test1.meta.yaml","content":"pack: default\nenabled: true\nrunner_type: orquesta\nname: test1\nentry_point: workflows/test1.yaml\nparameters:\n  arg1:\n    type: integer\n    default: 3\n"}]}
```

It doesn't seem to be a problem if I have it in a good state, and then edit messages and then save. Only after running execute.

I couldn't reproduce this behaviour in EWC 3.2 Workflow composer

PR #961, opened 2022-02-28
fix: Save workflow error when updating with parameters

Jappzy:
Fixes issue [862]( https://github.com/StackStorm/st2web/issues/862)

Saving a workflow immediately after creating and running it causes a 400 HTTP response code for a bad request. There is an extra property in the PUT request body that is not supported by the API. Finer details and replication steps can be found within the issue. These changes removes the `_name` property from the request body, resulting in a successful "save".

This also fixes some react errors regarding `ref`; it cannot be used for function components, and the `_details` variable is not defined either.

React Error Example (when loading multiple pages):
Screen Shot 2022-02-28 at 10 14 37 AM

### DIFF
--- a/apps/st2-actions/actions-panel.component.js
+++ b/apps/st2-actions/actions-panel.component.js
@@ -303,7 +303,6 @@ export default class ActionsPanel extends React.Component {
         </PanelView>
 
         <ActionsDetails
-          ref={(ref) => this._details = ref}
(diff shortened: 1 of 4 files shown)


amanda11:
This looks great to me, just needs a rebase to pick up latest from master.

I would like to give the changes a try before approval- hope to take it for spin later in week (unless someone else gets to review earlier than that).

Issue #492, opened 2018-01-25
Search operator not in operator dropdown for rules criteria


ahubl-mz:
When I added the [`search` operator](https://github.com/StackStorm/st2/pull/3833) to StackStorm, I didn't realize also needed to add it to the web interface (my bad ☹️). So now the `search` operator is supported in YAML files but doesn't show up in the list of operators in the web UI:

![screen shot 2018-01-25 at 12 12 37 pm](https://user-images.githubusercontent.com/32691407/35414124-7151bbd8-01d6-11e8-9af9-456f2a1c879d.png)

Adding this operator is not as straightforward as adding it to the `types` list in `st2web/modules/st2-criteria/criteria.component.js`, since its `pattern` isn't a simple static value, it's a list of subcriteria to match (see my PR above and the [documentation I wrote](https://docs.stackstorm.com/rules.html#advanced-comparison) for it to learn more), so it will have to be a special case.

Thanks!

Xorso:
So I have a few rules that utilize this feature and it looks like it is now breaking the webUI if you try to access that rule. This means I can't enable or disable the rule from the frontend or view it's criteria.

Here is an example of my rule:

```yaml
---
  name: 'deploy'
  pack: 'deploy'
  description: 'Config Ticket'
  enabled: false

  trigger:
      type: 'jira.event'

  criteria:
      trigger.event.body.issue_all.fields.issuetype.name:
        type: "iequals"
        pattern: "Config Console"
      trigger.event.body.changelog.items:
        type: "search"
        condition: any
        pattern:
          item.field:
            type: "iequals"
            pattern: "status"
          item.toString:
            type: "iequals"
            pattern: "Ready for Deploy "

  action:
    ...
```

The error I am getting is a react error visible from the developer console. The page goes completely white when I hit the error. This only happens when I go to actually view the rule in the UI.

```
Unhandled Promise Rejection: Invariant Violation: Minified React error #31; visit https://reactjs.org/docs/error-decoder.html?invariant=31&args[]=object%20with%20keys%20%7Bitem.toString%2C%20item.field%7D&args[]= for the full message or use the non-minified dev environment for full errors and additional helpful warnings.
```

I have tried a couple of different iterations to see if I could get it to work, but it looks like there is no support to handle this.

dylan-bitovi:
Bumping this issue, I am encountering this as well on ST3.5; my rule is even simpler than the one above:
```
  trigger.body.issue.fields.items:
    type: "search"
[...]

PR #964, opened 2022-03-07
Search Criteria Support in the Web UI

Jappzy:
Resolves: [492](https://github.com/StackStorm/st2web/issues/492)

Adds support for Search Criteria and nested parameter fields for rules.

This recording demonstrates the current application without the "search" option in the UI. Since search functionality is implemented in the back-end but not the front-end, navigating to a rule that uses search crashes the application.

https://user-images.githubusercontent.com/46300738/162856765-e75741b7-d062-4287-a315-042542fdb7d5.mov

This next recording demonstrates the changes: Added search criteria input fields, as well as not crashing the app when selecting a rule using search:

https://user-images.githubusercontent.com/46300738/162855561-76fe2216-f97f-4af3-a95a-f21bdea9db07.mov

### DIFF
--- a/modules/st2-criteria/criteria.component.js
+++ b/modules/st2-criteria/criteria.component.js
@@ -1,13 +1,13 @@
 // Copyright 2019 Extreme Networks, Inc.
 //
-// Licensed under the Apache License, Version 2.0 (the "License");
+// Licensed under the Apache License, Version 2.0 (the 'License');
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
 //
 //     http://www.apache.org/licenses/LICENSE-2.0
 //
 // Unless required by applicable law or agreed to in writing, software
-// distributed under the License is distributed on an "AS IS" BASIS,
+// distributed under the License is distributed on an 'AS IS' BASIS,
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
@@ -55,8 +55,8 @@ const types = {
 };
 
 const searchConditions = {
-  'any': "Any",
-  'all': "All",
+  'any': 'Any',
+  'all': 'All',
 };
 
 export default class Criteria extends React.Component {
@@ -80,7 +80,7 @@ export default class Criteria extends React.Component {
       ...data,
       [key]: {
         ...data[key],
-        condition
+        condition,
       },
     });
   }
@@ -93,12 +93,12 @@ export default class Criteria extends React.Component {
       [key]: {
         ...data[key],
         pattern: {
-          ...data[key]['pattern'],
+          ...data[key].pattern,
           ['item.data']: {
-            ...data[key]['pattern']['item.data'],
+            ...data[key].pattern['item.data'],
             pattern,
-          }
-        }
+          },
+        },
       },
     });
   }
@@ -111,12 +111,12 @@ export default class Criteria extends React.Component {
       [key]: {
         ...data[key],
         pattern: {
-          ...data[key]['pattern'],
+          ...data[key].pattern,
           ['item.data']: {
-            ...data[key]['pattern']['item.data'],
+            ...data[key].pattern['item.data'],
             type,
-          }
-        }
+          },
+        },
       },
     });
   }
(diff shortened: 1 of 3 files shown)


pradnya-orchestral:
@Jappzy , I have tested your changes, if we are selecting search from dropdwon, then condition is mandatory fields, so there should be some parameter field need to send while saving data in PUT API, so that we can get something for condition parameter in GET API. Here, need to check this because we are not getting anything for condition parameter in Get api so that on UI it is showing empty, I have made red square to highlight in following screenshot. Can you please
condition screen
 check this?

Jappzy:
@pradnya-orchestral Could you re-test? I added a screen recording and fixed the default values getting included in the PUT request.

mickmcgrath13:
@pradnya-orchestral are you able to re-review?

@Jappzy once things are good, we'll need to merge master -> this branch

nzlosh:
@pradnya-orchestral have you been able to test since @Jappzy applied the fixes for the problem you mentioned?

cognifloyd (approved):
From reading and based on the screen recording, this looks good to me. Without any more experienced JS reviewers, I'm inclined to merge this. If there are any issues, they should be found before release.

modules/st2-criteria/criteria.component.js
```diff
@@ -51,6 +51,12 @@ const types = {
   'timediff_gt': 'Later Than',
   'exists': 'Exists',
   'nexists': 'Doesn\'t Exist',
+  'search': 'Search',
+};
+
+const searchConditions = {
+  'any': 'Any',
+  'all': 'All',
 };
```
armab, line 60: There are more search conditions supported, according to https://docs.stackstorm.com/rules.html#advanced-comparison

```suggestion
const searchConditions = {
  'any': 'Any',
  'all': 'All',
  'any2any': 'Any2Any',
  'all2any': 'All2Any',
};
```

modules/st2-panel/panel.component.js
```diff
@@ -370,22 +370,37 @@ export class DetailsCriteriaLine extends React.Component {
...
   }
 
   render() {
-    const { name, type, pattern } = this.props;
+    const { name, type, pattern, condition } = this.props;
+
+    const search = type !== 'search' ? null : {
+      condition,
+      pattern: pattern['item.data'].pattern,
+      type: pattern['item.data'].type,
```
armab, line 383: Looks like there was a misunderstanding here and so this PR is always expecting to have `item.data` in the rule pattern.

I now see this is coming from the example: 
https://github.com/bitovi-stackstorm-exchange/st2-ui-workflow-tests/blob/main/rules/validate-alert.yaml#L16

But based on https://github.com/bitovi-stackstorm-exchange/st2-ui-workflow-tests/blob/4f576228aad17652ceb47b1f444f2624f4c116c8/rules/validate-alert.yaml#L21

instead of `item.data` could be `item.foo`, `item.bar` and so on, as that refers to the dictionary item.

And so based on implementation here, this works: 
```yaml
---
name: test_search1_ok
pack: st2
enabled: true

trigger:
  type: "core.st2.webhook"
  parameters:
    url: "webhook-alert/handle-alert"

criteria:
  trigger.body.alert:
    type: "search"
    condition: any
    pattern:
      item.data: # works
        type: "equals"
        pattern: "foo"

action:
  ref: "core.echo"
  parameters:
    message: "{{ trigger.body }}"
```

While this rule doesn't:
```yaml
---
name: test_search1_err
pack: st2
enabled: true

trigger:
  type: "core.st2.webhook"
  parameters:
    url: "webhook-alert/handle-alert"

criteria:
  trigger.body.alert:
    type: "search"
    condition: any
    pattern:
      item.time: # changed, it's just a dict item
        type: "equals"
        pattern: "foo"

action:
  ref: "core.echo"
  parameters:
    message: "{{ trigger.body }}"

```

armab (changes requested):
The feature implementation is missing the main point of rule with `search` criteria which could include multiple patterns.

And so this rule fails to render, despite being properly registered in st2: 
```yaml
---
name: test_search2_err
pack: st2
description: "Test search rule in web ui with multiple criteria patterns"
enabled: true

trigger:
  type: "core.st2.webhook"
  parameters:
    url: "webhook-alert/handle-alert"

criteria:
  trigger.body.alert:
    type: "search"
    condition: any
    pattern:
      item.data:
        type: "equals"
        pattern: "foo"
      item.time:
        type: "equals"
        pattern: "12:00"

action:
  ref: "core.echo"
  parameters:
    message: "{{ trigger.body }}"

```

https://docs.stackstorm.com/rules.html#advanced-comparison highlights more examples

PR #965, opened 2022-03-07
Auto-Save Changes in the Workflow Composer

Jappzy:
Closes #993

Adds a checkbox in the Workflow Composer to toggle "Auto-Save". When enabled, all changes in the Workflow Composer are debounced by 3 seconds and then automatically saved. This is an added enhancement to the save button and keyboard shortcut.

https://user-images.githubusercontent.com/46300738/163027284-9fb884ea-345d-4e80-9ebb-62a07d746c81.mov

### DIFF
--- a/modules/st2flow-details/index.js
+++ b/modules/st2flow-details/index.js
@@ -141,7 +141,7 @@ export default class Details extends Component<{
               );
             })
           }
-          <input id='autosave-checkbox' name='autosave-checkbox' type='checkbox' onChange={(e) => this.toggleAutosave(e.target.checked)} className={cx(style.autosave)} />
+          <input id='autosave-checkbox' name='autosave-checkbox' type='checkbox' onChange={(e) => { this.toggleAutosave(e.target.checked); onChange();}} className={cx(style.autosave)} />
           <label id='autosave-checkbox__label' htmlFor='autosave-checkbox' className={cx(style.autosave)}>Autosave</label>
           <ToolbarButton className={cx(style.code, 'icon-code')} selected={asCode} onClick={() => navigate({ asCode: !asCode })} />
         </Toolbar>
--- a/modules/st2flow-details/index.js
+++ b/modules/st2flow-details/index.js
@@ -112,7 +112,7 @@ export default class Details extends Component<{
   }
 
   render() {
-    const { actions, navigation, navigate } = this.props;
+    const { actions, navigation, navigate, onChange } = this.props;
 
     const { type = 'metadata', asCode } = navigation;
 
@@ -138,7 +138,7 @@ export default class Details extends Component<{
             asCode
               && <MetaEditor />
               // $FlowFixMe Model is populated via decorator
-              || <Meta />
+              || <Meta onChange={() => onChange()} />
           )
         }
         {
--- a/modules/st2flow-details/index.js
+++ b/modules/st2flow-details/index.js
@@ -126,6 +126,8 @@ export default class Details extends Component<{
 
     const { type = 'metadata', asCode } = navigation;
 
+    const { autosaveEnabled } = store.getState();
+
     return (
       <div className={cx(this.props.className, this.style.component, asCode && 'code')}>
         <Toolbar>
@@ -147,12 +149,13 @@ export default class Details extends Component<{
             <input
               id='autosave-checkbox' 
               name='autosave-checkbox' 
-              type='checkbox' 
+              type='checkbox'
               onChange={(e) => {
                 this.toggleAutosave(e.target.checked);
                 onChange();
               }} 
               className={cx(style.autosave)}
+              defaultChecked={autosaveEnabled}
             />
             <label id='autosave-checkbox__label' htmlFor='autosave-checkbox' className={cx(style.autosave)}>Autosave</label>
           </div>
--- a/modules/st2flow-details/index.js
+++ b/modules/st2flow-details/index.js
@@ -145,7 +145,7 @@ export default class Details extends Component<{
             id='autosave-checkbox' 
             name='autosave-checkbox' 
             type='checkbox' 
-            onChange={ (e) => { 
+            onChange={(e) => {
               this.toggleAutosave(e.target.checked);
               onChange();
             }} 
--- a/modules/st2flow-details/index.js
+++ b/modules/st2flow-details/index.js
@@ -91,6 +91,8 @@ export default class Details extends Component<{
     navigate: PropTypes.func,
 
     actions: PropTypes.array,
+
+    onChange: PropTypes.func,
   }
 
   sections = [{
--- a/modules/st2flow-details/index.js
+++ b/modules/st2flow-details/index.js
@@ -141,17 +141,21 @@ export default class Details extends Component<{
               );
             })
           }
-          <input 
-            id='autosave-checkbox' 
-            name='autosave-checkbox' 
-            type='checkbox' 
-            onChange={(e) => {
-              this.toggleAutosave(e.target.checked);
-              onChange();
-            }} 
-            className={cx(style.autosave)}
-          />
-          <label id='autosave-checkbox__label' htmlFor='autosave-checkbox' className={cx(style.autosave)}>Autosave</label>
+          <div
+            style={{display: 'flex'}} title="Automatically save the workflow on every change"
+          >
+            <input
+              id='autosave-checkbox' 
+              name='autosave-checkbox' 
+              type='checkbox' 
+              onChange={(e) => {
+                this.toggleAutosave(e.target.checked);
+                onChange();
+              }} 
+              className={cx(style.autosave)}
+            />
+            <label id='autosave-checkbox__label' htmlFor='autosave-checkbox' className={cx(style.autosave)}>Autosave</label>
+          </div>
           <ToolbarButton className={cx(style.code, 'icon-code')} selected={asCode} onClick={() => navigate({ asCode: !asCode })} />
         </Toolbar>
         {
--- a/modules/st2flow-details/index.js
+++ b/modules/st2flow-details/index.js
@@ -141,14 +141,23 @@ export default class Details extends Component<{
               );
             })
           }
-          <input id='autosave-checkbox' name='autosave-checkbox' type='checkbox' onChange={(e) => { this.toggleAutosave(e.target.checked); onChange();}} className={cx(style.autosave)} />
+          <input 
+            id='autosave-checkbox' 
+            name='autosave-checkbox' 
+            type='checkbox' 
+            onChange={ (e) => { 
+              this.toggleAutosave(e.target.checked);
+              onChange();
+            }} 
+            className={cx(style.autosave)}
+          />
           <label id='autosave-checkbox__label' htmlFor='autosave-checkbox' className={cx(style.autosave)}>Autosave</label>
           <ToolbarButton className={cx(style.code, 'icon-code')} selected={asCode} onClick={() => navigate({ asCode: !asCode })} />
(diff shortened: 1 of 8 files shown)


amanda11:
Thanks for the contribution.

I'm wondering if its possible to make this something you can switch on when you want. Often I don't want changes automatically saved when I'm editing workflows as I might be experimenting, or showing how a workflow could be altered - so to have it automatically save is NOT what I would want it to do.

So I think this would need to be something that you enable either on a system-wide or per edit of a workflow.

mickmcgrath13:
Needs rebase/merge from `master`

Jappzy:
A toggle to enable/disable autosave is a great idea! I can add that in the workflow composer and rebase.

amanda11:
> A toggle to enable/disable autosave is a great idea! I can add that in the workflow composer and rebase.

Thanks. I think it's safest to have disabled by default,as if you are editing a workflow that is called by a rule - and say making a couple of changes - auto save would make them live part way through - which could be bad.

But if you are editing a brand new workflow or one not live - then you may want to enable auto save. 

Probably safest to make auto save be something the user makes conscious choice to enable.

amanda11 (changes requested):
I would like to see the auto-save be switched off by default, but something that when editing a workflow the user can switch on. 
As if this is a live workflow - auto-save would cause issues as it would mean the workflow was changed whilst in middle of making a set of changes. 
However if it's a non-live workflow then auto-save could be really useful, but should be something the user enables each time.

I will wait to test how it works until its a toggle, as I that will increase the amount of tests to run.

I'd also be useful for some feedback from users as to whether 25 seconds for the autosave is going to be too much or too little. If you are editing a workflow then alot of the time it could be in an invalid state. So wondering if 25 seconds is too frequent?

I'd also like to see some information added to the PR as to when the auto-save kicks in, is it every 25 seconds regardless of whether I'm editing, only if I've done an edit, only when I move out of a field?
How does the Undo/Redo work in regard to auto-save? If in Auto-Save mode then is Undo/Redo redundant, or what does it undo until - the last fixed Save time, or the last auto-save position?

Jappzy:
@amanda11 Thanks for your feedback! I added a toggle for the autosave functionality with a debounce time of 3 seconds. It's indifferent to the undo/redo functionality; clicking "undo" after an autosave will not revert the save, it will apply the next change in the undo stack, unaffected by the save.

rush-skills:
I feel that this feature would be more useful if it is opt-in (user can toggle auto-save to ON in the workflow composer) and then it is rather instantaneous (1 sec?) rather than a random wait (3-25 seconds?), since the wait might make the experience inconsistent (I made changes thinking they will be autosaved but closed the tab before the autosave triggered).

rush-skills:
I feel that this feature would be more useful if it is opt-in (user can toggle auto-save to ON in the workflow composer) and then it is rather instantaneous (1 sec?) rather than a random wait (3-25 seconds?), since the wait might make the experience inconsistent (I made changes thinking they will be autosaved but closed the tab before the autosave triggered).

mickmcgrath13:
@rush-skills I agree it should be opt-in. @Jappzy can you confirm whether or not the latest updates are "off by default and opt-in" ?

As far as the debounce, I think we want a small wait so that it's not overloading the server with api calls due to a request for every .  1-2 second might be okay.  @Jappzy and @amanda11 , thoughts on dropping the timeout to 1s?

Best case scenario would be to make it configurable via the st2config, but maybe that's work for a future ticket as it'd be more effort.

Jappzy:
@amanda11 @rush-skills @mickmcgrath13  Can you elaborate on the repeating requests every 1 second? This PR has the autosave feature off by default and debounces all of the change requests for 3 seconds after each change. There are no requests every second and it only updates when something changed. There is a screen recording demo in the description showcasing the functionality.

mickmcgrath13:
@Jappzy That's How I understood it:

- auto-save off by default
- toggle auto-save on, and
  - making a change causes a timer for 3 seconds - after which the save is triggered
  - additional changes that happen within the 3 seconds resets the timer (i.e. debounce)

What I was referring to is essentially to reset the timer after 1s vs 3s.

mickmcgrath13:
@amanda11 does that alleviate your concerns?

rush-skills:
@Jappzy  If it is off-by-default and does not send a request unless the workflow is actually changed, it is good to me. 
For the 1s vs 3s question - I don't think it changes any thing on the server wrt load (does 1 sec mean more writes during the editing phase? is that problematic?), and I will prefer 1 second over 3 seconds just because it seems more UX friendly (I will be less likely to face a situation where I turn auto-save on, make changes, close the tab soonish, and don't have the changes persisted because it was less than 3 secs).

amanda11:
> @amanda11 does that alleviate your concerns?

Yes the design sounds great. I've not had a chance to install and have a play to test it out. But design sounds good, and am ok with the 1second proposal as well.

Jappzy:
@amanda11 great! I changed the debounce timer to 1 second so it should be good to test now. Thanks for the review!

amanda11:
I noticed some oddities:

Doesn't auto-save if you have unsaved changes before click on auto-save.
--------------------------------------------------------------------------------------------------
1. Create new workflow
2. Give it a name, Hit Save
3. Then the description in the meta-data
4. Then click on auto-save

The unsaved changes in the description added at step3 don't get auto-saved. It only saves them if I make another change.

Doesn't auto-save if making changes in the YAML entry input
---------------------------------------------------------------------------------
If you switch to the edit yaml directly when in the auto-save, then changes you make aren't autosaved

1. Create new workflow
2. Give it a name and description, Hit Save
3. Then enable auto-save
4. Use the <> to go into the direct input buttons. When you do this then the changes I make aren't auto-saved.

After I'd done this second one, then if I opened new workflows the auto-save seemed to stop working at all....

amanda11 (changes requested):
I noticed some oddities:

Doesn't auto-save if you have unsaved changes before click on auto-save.
--------------------------------------------------------------------------------------------------
1. Create new workflow
2. Give it a name, Hit Save
3. Then the description in the meta-data
4. Then click on auto-save

The unsaved changes in the description added at step3 don't get auto-saved. It only saves them if I make another change.

Doesn't auto-save if making changes in the YAML entry input
---------------------------------------------------------------------------------
If you switch to the edit yaml directly when in the auto-save, then changes you make aren't autosaved

1. Create new workflow
2. Give it a name and description, Hit Save
3. Then enable auto-save
4. Use the <> to go into the direct input buttons. When you do this then the changes I make aren't auto-saved.

After I'd done this second one, then if I opened new workflows the auto-save seemed to stop working at all....

mickmcgrath13:
@Jappzy Can you take a look at @amanda11's latest comment?

WestonVincze:
Thanks for the detailed feedback @amanda11 . 

I've fixed the first issue by adding a call to the save function when the autosave checkbox is turned on. I also added the autosave functionality to the browser editors (metadata and workflows).

amanda11:
@Jappzy I had some time to test today, but unfortunatley I can't download the package as the CircleCI tests are failing.
We need to look at why we are getting incompatible versions of node being used for the CircleCI runs.
It's not particular to your PR its failing on all st2web PRs, but if you have a chance to look that would be great.

amanda11:
think I've found the problem with st2web, and I'm looking at a PR to fix.

amanda11:
@Jappzy Can you look at why the lint checks are failing on this PR? I'd like to get it so we've got a built instance I can install on a test box. Thanks.

WestonVincze:
@amanda11 fixed the CI issues. You should be able to build it now.

nzlosh:
@amanda11 Have your review requirements been satisfied?

amandam11:
> @amanda11 Have your review requirements been satisfied?

I missed the comment that lint had been fixed, I've not had a chance to re-test. I'll try and do that this week.

amanda11:
> @amanda11 Have your review requirements been satisfied?

I missed the lint had been fixed, I'll try and install and test this week.

amanda11 (changes requested):
I am still seeing some cases where changes aren't being autosaved. Although the items I mentioned before are fixed, there are new problems.

Changing colour of transition doesn't auto-save
---------------------------------------------------------------
1.  Without auto-save off, create a workflow with 2 actions and a transition
2.  Switch on auto-save
3. Look at the workflow file created
4. Switch on auto-save
5. Change the colour of the transition 
6. Look at the workflow file - it hasn't got saved

Similarly if I go into the condition on the transition, and enter in the "when" box it doesn't get saved.

I can tell that the auto-save is working, as if I move a box then it does auto-save.

Can we go through everything in the UI and check that it does auto-save everything?

I found another thing if I use the publish on the transition, and in variables to publish, again these don't get saved.

Not auto-saving when I'm adding actions
-----------------------------------------------------
I'm not sure when its deciding to trigger the save now.  As its not donig it when I drag actions or update them in the workflow

1. I created a new workflow, entered in the name. Saved, and then put on auto-save
2. I entered in a description, it did auto-save this
3. I dragged an action in - nothing got auto-saved
4. I added in value to the action's parameter, nothing got auto-saved
5. I moved the action - then it got auto-save...

amanda11:
@Jappzy @WestonVincze I've found different problems with the auto-save now, please see the review comments. Worth double-checking after any change that its auto-saving when you change anything in the UI - the list of ones where it isn't, might not be exhaustive.

armab:
@amanda11 This PR had some changes and is ready for your review again.
I'll test it too, but you may know the previous places where it didn't work as expected.
Thanks for your feedback!

amanda11:
> @amanda11 This PR had some changes and is ready for your review again.
> 
> I'll test it too, but you may know the previous places where it didn't work as expected.
> 
> Thanks for your feedback!
> 
> 

Thanks @armab - should have time later in week to test this, and I will report back on ticket

armab (commented):
I've tested it at a high level and it worked ok to me.

@cded As this PR is awaiting more reviews, perhaps one minor enhancement proposal from me would be adding a tooltip for the "Autosave" text so when the user is hovering his cursor over the "Autosave", it shows a small hint description as we do for other buttons:
image


image

Something along the lines, but feel free to improve if you see a better description: 
> Automatically save the workflow on every change, with a 1s delay

modules/st2flow-details/index.js
```diff
@@ -131,20 +141,35 @@ export default class Details extends Component<{
               );
             })
           }
+          <div
+            style={{display: 'flex'}} title="Automatically save the workflow on every change"
```
armab, line 145: 👍

armab (approved):
LGTM

Will wait for another  👀 by @amanda11

amanda11 (changes requested):
Found a problem:

1. If I am on the UI entering mode, and I use the Parameters tab to add a Parameter. If I add the parameter and do Done, then the metadata and main file are not updated for the new parameter. Only when I do another change such as add a transition does it save the fact that I've added a parameter.
2. Similar problem if delete Parameter or edit Parameter.

These were the only problems I found.

amanda11:
@mickmcgrath13 @armab @cded I just found one scenario where it's not autosaving - see review comment.

cded:
@amanda11 Thanks for the catch, I updated the PR to handle auto save on adding, editing and deleting parameters. I also fixed an issue with the auto save checkbox where toggling the details panel didn't save the checkbox state.

amanda11 (approved):
LGTM - just need to check with release manager about when to merge

armab:
@amanda11 Thanks for all the reviews 👍 
Based on Slack, looks like we're good to merge, - appreciate the follow-up with @nzlosh!

Issue #981, opened 2022-06-30
@stackstorm/module-api does not handle additional data in ContentType headers (ie: content-type: application/json; charset=utf-8)


luislobo:
Our production Nginx configuration has set `charset utf-8;` making `Content-Type` header to be `application/json; charset=utf-8`, not just `application/json`.

`@stackstorm/module-api` compares `headers['content-type']` to be exactly `application/json`.

I'll be creating a PR that fixes this problem.

PR #982, opened 2022-06-30
Handle cases where `content-type` is not only `application/json`

luislobo:
Fixes https://github.com/StackStorm/st2web/issues/981

Some comments about the request function:

- It seems like it's been migrated from st2client.js
- @key10 and I reviewed it and it seems like there is an opportunity to improve it

Some possible improvements/questions:
1. Should `raw` return the data without doing any JSON.parse? If so, then `transformResponse` should not be used.
2. On line 234, `body` is assigned `data`. Assuming `data` is a string at this moment, `body` is not a pointer of `data`, but rather a copy. Then, the changes done on `body` on lines 244-248 (for parsing JSON) are superfluous, since the function is returning `data`.
3. Line 232 could be done before 229, and not have to repeat defaulting to `{}` on both 229 and 230.

Let us know about those comments/questions and we can apply changes as needed.


luislobo:
@amanda11 @armab 

I'd appreciate if you can check this one out when you have a chance.

Thank you

modules/st2-api/api.js
```diff
@@ -192,7 +192,8 @@ export class API {
       params: query,
       headers,
       transformResponse: [ function transformResponse(data, headers) {
-        if (typeof data === 'string' && headers['content-type'] === 'application/json') {
+        if (typeof data === 'string' && headers['content-type'] &&
+            headers['content-type'].indexOf('application/json') >-1) {
```
amanda11, line 196: Minor comment, I'd suggest a space before the -1.

amanda11 (commented):
Changes look ok to me, but we need to fix the CircleCI failures - which I don't think are related to this PR in particular.
I added one minor comment on spacing, but looks ok, but we need to see if we can fix the CircleCI problems. It looks like a problem with version of node.

amanda11:
think I've found the problem with st2web, and I'm looking at a PR to fix.

amanda11:
@luislobo If you can merge in latest master this should hopefully build ok now.

luislobo:
Sure, will do.

modules/st2-api/api.js
luislobo: Done :)

luislobo:
Seems like tests passed now :)

PR #989, opened 2022-09-26
Updated changelog

nzlosh:


cognifloyd (approved):

CHANGELOG.rst
```diff
@@ -1,13 +1,37 @@
...
+
+Changed
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+
+Removed
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+
+v2.4.3
```
armab, line 36: The Changelog is a good addition, but the reality st2web version was last updated 3 years ago: https://github.com/StackStorm/st2web/commit/656c1e49ca6b4bd9580a105d62f6ee5aa2980600

In other words, as a user, there is no easy way to find out which change shipped with which stackstorm version.
Does it make sense to point or hint to the respective StackStorm version instead? As that's what really matters.

PR #1009, opened 2023-10-20
Update dependencies

enykeev:
This PR updates softer defined \ transient dependencies and is supposed to cover for a number of open dependabot PRs in one feel swoop.

Three minor issues discovered:
- Axios no longer need data to be set on a request for `content-type` to be populated. Instead, it now sends the value provided in the `data` field verbatim and breaks auth flow. Removing previous hack fixes the problem.
- `fontello-update`'s dependency `open` has been defined very broadly and as `>=6.0.0` and switched to ESM starting 9.0.0. We do not use ESM on this project and it isn't worth to start because of a single transient dependency.
- When adding `ST2_PROTOCOL` we forgot to add it to our `serve` task. This creates a problem when you trying to run local version of st2web against dockerized st2.

In general, I have to admit, I'm beyond impressed with how well this project held up for all this years with minimal maintenance it had. Thanks to the community for keeping it afloat.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,7 +6,7 @@ in development
 
 Changed
 ~~~~~~~
-* Updated various dependencies. #1009
+* Updated various dependencies (security). #1009
 
   Contributed by @enykeev
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,12 @@ Changelog
 in development
 --------------
 
+Changed
+~~~~~~~
+* Updated various dependencies. #1009
+
+  Contributed by @enykeev
+
 Fixed
 ~~~~~
 * Fixed CircleCI tests by pinning lerna@6.0.0. #1008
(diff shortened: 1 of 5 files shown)


armab (changes requested):
Thanks a lot!

Could you please also update the https://github.com/StackStorm/st2web/blob/master/CHANGELOG.rst as well?

Besides that, all good.
A note for myself before merging: I'll probably need to do a quick st2web visual check to make sure nothing broke because of updates.

CHANGELOG.rst
```diff
@@ -4,6 +4,12 @@ Changelog
 in development
 --------------
 
+Changed
+~~~~~~~
+* Updated various dependencies. #1009
```
armab, line 9:
```suggestion
* Updated various dependencies (security). #1009
```

armab (approved):
I quickly ran a few visual checks on updated st2web and found no regressions.

Thanks for updating the dependencies! 👍

Kami (approved):

armab:
@Kami welcome to the party! 🎉

PR #1020, opened 2023-10-28
Update additional dependencies

enykeev:
- `moment` is reported by Snyk as vulnerable and an be safely updated to the next minor version
- `lodash.template` is reported by Snyk as vulnerable. It's also not being used anywhere in code and can be safely deleted.
- `react-textarea-autosize` seem to be breaking CI by attempting to evaluate global variable that is not available in all environments. It can not be updated to the latest version, but it could safely updated to previous major version.

I've also attempted to update `axios` as it's also reported as being vulnerable by Snyk, but the package transitioned to using ESM and is incompatible with our current build system.


armab (commented):
Thanks!

Please also mention it in the Changelog

armab (approved):

cognifloyd (approved):

PR #1032, opened 2025-01-15
Add Ubuntu Jammy to package build/deploy

nzlosh:


amanda11 (changes requested):
There is still a mention of bionic when it gets package version:
```
 PKG_RELEASE=$(packagecloud.sh next-revision bionic ${PKG_VERSION} st2web)

nzlosh:
Good spotting, thank you.  I've removed the hard coded reference to `bionic` and the script will now use the first definition from the list of distros defined in `RPM` and `DEB`.

amanda11 (approved):

cognifloyd (approved):

Issue #1005, opened 2023-07-19
passwords visible in preview


fdrab:
Hello,

I've found past issue (https://github.com/StackStorm/st2web/issues/411) that should have solved this, but it seems in 3.8.0 the preview leaks fields marked as secret:

Do I have to configure something in the st2.conf? Or is this by design?

![Screenshot 2023-07-19 173418](https://github.com/StackStorm/st2web/assets/118189175/441a140b-f47b-4aad-8552-83567f720b35)

BR,
Filip

armab:
This sounds like a bug indeed as secrets should be masked. Thanks for the report.

If someone is interested in contributing, the fix should be done in the st2 core which provides st2web with an API response.

docbyte86:
Same issue while checking the past executions in the execution tab.

fdrab:
> Same issue while checking the past executions in the execution tab.

can you post example screenshot? I see secrets properly masked in past execution outputs:
![Screenshot 2023-07-27 084430](https://github.com/StackStorm/st2web/assets/118189175/984769d7-d769-4919-a92a-7cc1b805f35a)

PR #1036, opened 2025-07-22
sanitize preview when handling secret parameters

fdrab:
re-introducing sanitization of preview when handling secrets - issue #1005

### DIFF
--- a/apps/st2-actions/actions-details.component.js
+++ b/apps/st2-actions/actions-details.component.js
@@ -17,6 +17,7 @@ import React from 'react';
 import { PropTypes } from 'prop-types';
 import { connect } from 'react-redux';
 import store from './store';
+import _ from 'lodash';
 
 import api from '@stackstorm/module-api';
 import notification from '@stackstorm/module-notification';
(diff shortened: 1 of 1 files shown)


nzlosh (approved):

fdrab:
I'll add it.

cognifloyd:
With this change the only way to check i typed a secret correctly is to use the visibility button on each secret field added in #1035 ?

Will this affect secret fields with an object or array? Today, preview is the only way I can validate that I typed or pasted structured data correctly.

cognifloyd:
Does the preview make `secret_object` look like it is a string? If so, could it be masked as something like `{****}` instead of `"****"`?

fdrab:
like this?
image

fdrab:
If the above is not satisfactory, perhaps I can, instead of replacing all chars with * just generate random alphanumeric strings to replace the originals with? Would that be better?

cognifloyd:
That is better. Something nicer would be great, but it's probably not required to get this in. That would be a great follow-up PR though.

I think we need the hide/show secrets button merged before this goes in.

PR #1045, opened 2026-07-29
Fix for issue #364

TimothyDJones:
Instead of checking each input parameter for `'********'` (as in earlier implementation), we create a copy of the *entire* payload, then iterate over each element of the payload. If the parameter is *NOT* `'********'`, then simply use that value; otherwise, if the original value was blank/null, then use that value or delete that parameter altogether.


guzzijones (approved):

guzzijones:
This is a good fix.  I have also noticed in the gui when you rerun an action with secret parameters it grabs the invalid ****** instead of leaving it to use the default.  It is most annoying with core.remote where we have a private key path as a parameter secret.

